PR #33, opened 2020-05-21
integrate go-coap/v2 to cloud

jkralik:


http-gateway/service/service.go
```diff
@@ -95,7 +95,8 @@ func (s *Server) Serve() error {
 
 // Shutdown ends serving
 func (s *Server) Shutdown() error {
-	for _, v := range s.requestHandler.manager.observations {
+	observations := s.requestHandler.pop()
```
cizmazia, line 98: Is there a timeout for closing observations to prevent the shutdown from hanging?

cizmazia (approved):

http-gateway/service/service.go
jkralik: `func (s *subscribeSession) OnClose() ` is a function make action without any sync with peers.

PR #144, opened 2021-01-08
Move cqrs repository to resource-aggregate/cqrs

jkralik:


coap-gateway/coapconv/coapconv.go
```diff
@@ -161,12 +161,12 @@ func MakeCommandMetadata(sequenceNumber uint64, connectionID string) pbCQRS.Comm
 	}
 }
 
-func MakeConfirmResourceRetrieveRequest(deviceID, href, correlationId string, authCtx pbCQRS.AuthorizationContext, connectionID string, req *pool.Message) pbRA.ConfirmResourceRetrieveRequest {
+func MakeConfirmResourceRetrieveRequest(deviceID, href, correlationId string, authCtx *pbCQRS.AuthorizationContext, connectionID string, req *pool.Message) pbRA.ConfirmResourceRetrieveRequest {
```
ondrejtomcik, line 164: We're currently working only with devices, but the plan is to host resources also by other services. In such a scenario, resources are published by service and not by the device. What do you think about making this transparent in our naming? Renaming everywhere deviceId to e.g. `resourceServerID `? I wouldn't do it in this MR because it will blow up, but I would do it in v2.

resource-aggregate/cqrs/aggregate/aggregate.go
```diff
@@ -0,0 +1,243 @@
...
+// Command user defined command that will handled in AggregateModel.HandleCommand
+type Command = interface{}
+
+// AggregateModel user model for aggregate need to satisfy this interface.
+type AggregateModel = interface {
+	eventstore.Model
+
+	HandleCommand(ctx context.Context, cmd Command, newVersion uint64) ([]eventstore.Event, error)
+	TakeSnapshot(version uint64) (snapshotEvent eventstore.Event, ok bool)
+	GroupId() string //defines group where model belows
```
ondrejtomcik, line 21: We don't have groups, they were removed. Or what does this group represents?

resource-aggregate/cqrs/aggregate/aggregate.go
```diff
@@ -0,0 +1,243 @@
...
+		return time.Now().Add(time.Millisecond * 10), nil
+	}
+}
+
+// FactoryModelFunc creates model for aggregate
+type FactoryModelFunc = func(ctx context.Context) (AggregateModel, error)
+
+// Aggregate holds data for Handle command
+type Aggregate struct {
+	groupID             string
```
ondrejtomcik, line 44: What represents the groupID?

resource-aggregate/cqrs/eventbus/event.go
```diff
@@ -0,0 +1,31 @@
...
+	Version() uint64
+	EventType() string
+}
+
+//EventUnmarshaler provides event.
+type EventUnmarshaler = interface {
+	Version() uint64
+	EventType() string
+	AggregateID() string
+	GroupID() string
```
ondrejtomcik, line 18: Same here.

coap-gateway/coapconv/coapconv.go
jkralik: too long name but I'm not against it.

resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
```diff
@@ -58,68 +91,706 @@ func NewEventStore(config Config, goroutinePoolGo eventstore.GoroutinePoolGoFunc
...
+	}
+	if len(errors) > 0 {
+		return fmt.Errorf("%+v", errors)
+	}
+	return nil
+}
+
+// LoadUpToVersion loads aggragates events up to a specific version.
+func (s *EventStore) LoadUpToVersion(ctx context.Context, queries []eventstore.VersionQuery, eh eventstore.Handler) error {
+	s.LogDebugfFunc("mongodb.Eventstore.LoadUpToVersion start")
```
ondrejtomcik, line 491: We should get rid of these logs. Probably not yet, but definitely before merging.

resource-aggregate/cqrs/aggregate/aggregate.go
jkralik: GroupID = DeviceID so we can renamed them to `resourceServerID`

ondrejtomcik (approved):
There are few renamings that need to be done, but not within this MR.
groupID -> resourceServerID
deviceID -> resourceServerID
aggregateID -> resourceID
get rid of debug logs printing command execution times. We will use open telemetry for this.

ondrejtomcik:
This issue https://github.com/plgd-dev/cloud/issues/149 needs to be resolved before merging branch v2 to master.

PR #153, opened 2021-01-18
http-gateway: expose react app

jkralik:


bundle/Dockerfile
```diff
@@ -48,6 +48,14 @@ RUN mkdir -p ./nats-server
 RUN unzip ./nats-server.zip -d ./nats-server
 RUN cp ./nats-server/*/nats-server /go/bin/nats-server
 
+FROM ubuntu:20.04 AS build-web
```
ondrejtomcik: Why you are not using node container?

bundle/run.sh
```diff
@@ -247,6 +247,7 @@ Dial:
 JwksURL: ${JWKS_URL}
 ResourceDirectoryAddr: ${RESOURCE_DIRECTORY_ADDRESS}
 CertificateAuthorityAddr: ${CERTIFICATE_AUTHORITY_ADDRESS}
+WWWDirectory: /usr/local/var/www
```
ondrejtomcik: This should be the default value in the http-gw service.

http-gateway/Dockerfile
```diff
@@ -2,7 +2,16 @@ FROM cloud-build AS build
 WORKDIR $GOPATH/src/github.com/plgd-dev/cloud/http-gateway
 RUN go build -o /go/bin/service ./cmd/service
 
+FROM ubuntu:20.04 AS build-web
```
ondrejtomcik: Use node container.

http-gateway/service/httpApi.go
```diff
@@ -120,6 +117,11 @@ func NewHTTP(requestHandler *RequestHandler, authInterceptor kitHttp.Interceptor
 	r.HandleFunc(uri.WsStartDevicesObservation, requestHandler.startDevicesObservation).Methods(http.MethodGet)
 	r.HandleFunc(uri.WsStartDeviceResourcesObservation, requestHandler.startDeviceResourcesObservation).Methods(http.MethodGet)
 
+	// serve www directory
+	if requestHandler.config.WWWDirectory != "" {
```
ondrejtomcik: UI will be always embedded in the http gateway container. It's more clean to use boolean value to set if the UI should be hosted or not.

ondrejtomcik (changes requested):

bundle/Dockerfile
jkralik: Fixed

bundle/run.sh
jkralik: Fixed

http-gateway/Dockerfile
jkralik: Fixed

http-gateway/service/config.go
```diff
@@ -9,14 +9,16 @@ import (
...
+	Address                  string
+	Listen                   certManager.Config
+	Dial                     certManager.Config
+	JwksURL                  string
+	ResourceDirectoryAddr    string
+	CertificateAuthorityAddr string
+	WebSocketReadLimit       int64
+	WebSocketReadTimeout     time.Duration
+	UIEnabled                bool
+	UIDirectory              string
```
ondrejtomcik, line 21: What is the use case to have it configurable ?

jkralik: I don't want to hard-code it, but I don't have use case for now. So I want to leave as is.

ondrejtomcik (approved):

PR #159, opened 2021-01-26
[ui] Table component and useApi fetch hook

kuby:
- Table component with sort and pagination
- useApi hook
- Things table


ondrejtomcik (approved):

http-gateway/web/src/components/table/table.js
```diff
@@ -0,0 +1,140 @@
...
+          />
+        </div>
+      )}
+    </>
+  )
+}
+
+Table.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
+  columns: PropTypes.arrayOf(PropTypes.object).isRequired,
```
oskarbauer, line 125: Probably should make a shape for this.

oskarbauer (approved):

http-gateway/web/src/components/table/table.js
kuby: Will do, in the next version :) thank

PR #161, opened 2021-01-27
[UI] Label component and device details page

kuby:


http-gateway/web/src/containers/things/things-list-page.js
```diff
@@ -26,7 +26,7 @@ export const Things = () => {
         Header: _(t.name),
         accessor: 'device.n',
         Cell: ({ value, row }) => (
-          <Link to={`/things/${row.values.id}`}>{value}</Link>
+          <Link to={`/things/${row.values?.['device.di']}`}>{value}</Link>
```
oskarbauer: is "di" correct?  If it is, I think it should be deviceIndentfier or something else like that.   id is clear, di is not

oskarbauer (dismissed):
device.n and devide.di are not clear to me.

ondrejtomcik:
The di represents device id and n device name. Based on OCF spec.

ondrejtomcik (approved):

http-gateway/web/src/containers/things/things-list-page.js
kuby: That's how it comes from the API (`di` stands for device ID), I can't do much about it :)

oskarbauer (approved):

PR #167, opened 2021-02-03
[ui] Things view modal with retrieve

### LINKED ISSUE #368: Read & Update resources on plgd Dashboard
Being able to retrieve and update device resources using plgd Dashboard.

Easy to use.

None

kuby:
- Modal component
- Button component
- Dropdown component
- Resource details view modal with retrieve functionality


http-gateway/web/src/containers/things/things-details-page.js
```diff
@@ -48,17 +56,78 @@ export const ThingsDetailsPage = props => {
...
+    }
+
+    setLoadingResource(true)
+
+    try {
+      const interfaceGetParam = currentInterface
+        ? `interface=${currentInterface}`
+        : ''
+      const {
+        data: { if: ifs, rt, ...resourceData }, // exclude the if and rt
```
oskarbauer, line 76: in this case I would give it a better name than ifs.

http-gateway/web/src/containers/things/_things-resources-update-modal.js
```diff
@@ -0,0 +1,146 @@
...
+
+ThingsResourcesUpdateModal.propTypes = {
+  onClose: PropTypes.func,
+  data: PropTypes.shape({
+    di: PropTypes.string.isRequired,
+    href: PropTypes.string.isRequired,
+    types: PropTypes.arrayOf(PropTypes.string),
+    interfaces: PropTypes.arrayOf(PropTypes.string),
+  }),
+  resourceData: PropTypes.object,
```
oskarbauer, line 136: as always, a shape would be nice. But in this case if it's not possible then eslint disable it.

http-gateway/web/src/containers/things/things-details-page.js
oskarbauer: Looks like ifs is not even used though

oskarbauer (approved):

http-gateway/web/src/containers/things/things-details-page.js
kuby: This is this parameters official name. The name change I did here from "if" to "ifs" is only because `if` is a reserved keyword for conditions.

http-gateway/web/src/containers/things/_things-resources-update-modal.js
kuby: Not possible, it can be literally anything. Also, new eslint does not force you to disable it, it is not throwing an error nor a warning.

http-gateway/web/src/containers/things/things-details-page.js
kuby: Yes, it is not used, nor the `rt`, as the comment at the end of line states, it is only for excluding them from the object.

PR #180, opened 2021-02-23
coap-gateway: update device status online in shortime than exp token

jkralik:


rafajpet (approved):

coap-gateway/service/config.go
```diff
@@ -7,6 +7,15 @@ import (
 	"github.com/plgd-dev/kit/sync/task/queue"
 )
 
+type DeviceStatusValidityConfig struct {
```
ondrejtomcik: `DeviceStatusExpiration`, + please document the behaviour to the md.

coap-gateway/service/config.go
```diff
@@ -7,6 +7,15 @@ import (
...
+	Enable   bool          `yaml:"enable" json:"enable"`
```
ondrejtomcik: we're using past tense - `Enabled`

coap-gateway/service/config.go
```diff
@@ -7,6 +7,15 @@ import (
...
+	Validity time.Duration `yaml:"validity" json:"validity"`
```
ondrejtomcik: `ExpiresIn`

ondrejtomcik: Is this value somehow related to ping?
e.g. if user configures expiresIn to 5secodns but ping between device and cloud is set to 30 seconds?

coap-gateway/service/config.go
ondrejtomcik: There is still pending MR from @Trishia . Is this going to create a conflict? Shouldn't we firstly merge Trishias changes?

coap-gateway/service/config.go
jkralik: no it is not related to ping

ondrejtomcik: And shouldn't be? at least, it should be documented how it behaves with combination with ping.

coap-gateway/service/config.go
jkralik: Done

coap-gateway/service/config.go
jkralik: Done

coap-gateway/service/config.go
jkralik: There is lot's of work to finish it - mainly in tests. So I don't want to wait for it. At first @ondrejtomcik pls validate yaml files and than when you are ok with them I can fix tests there.

ondrejtomcik (approved):

PR #184, opened 2021-02-25
[ui] WebSocket implementation

kuby:
- Websocket client and Websocket manager
- Event emitter
- Live device status handling
- Toast notifications with redirects
- Unregistered state handling
- Switch component
- Documentation updates
- Pagination bugfix
- Table case insensitive sorting


http-gateway/web/src/common/services/emitter.js
```diff
@@ -0,0 +1,50 @@
+class _Emitter {
+  constructor() {
+    this.events = {}
+  }
+
+  // Register an event listener
+  on(event, listener) {
+    if (typeof this.events[event] !== 'object') {
```
oskarbauer, line 8: This is a common pattern you are using. A helper util function might be useful.

http-gateway/web/src/common/services/init-services.js
```diff
@@ -0,0 +1,21 @@
...
+export const InitServices = () => {
+  useEffect(() => {
+    // Register the default WS instances
+    if (!WSManager.isInitialized) {
+      WSManager.addWsClient(thingsWSClient)
+      WSManager.registerWSClients()
+    }
+
+    return () => {
+      // WSManager.unregisterWSClients()
```
oskarbauer, line 16: This is commented off, is it handled somewhere else?

http-gateway/web/src/common/services/ws-manager.js
```diff
@@ -0,0 +1,206 @@
...
+    }
+  }
+
+  /**
+   * Removes all WS clients which includes the name (or part of the name) given in the partialName argument.
+   * @param {String} partialName - Partial or full name of the WS client.
+   */
+  removeAllByPartialNameFromWsClient = partialName => {
+    Object.keys(this.ws)
+      .filter(key => key.indexOf(partialName) !== -1)
```
oskarbauer: use the native includes function.

http-gateway/web/src/common/services/ws-manager.js
```diff
@@ -0,0 +1,206 @@
...
+        if (this.ws[id] && listener) {
+          this.ws[id].connect(delayMessage)
+          this.ws[id].onMessage = listener
+
+          // Reconnect on close after "this.reconnectTimeMs" seconds
+          this.ws[id].onClose = () => {
+            if (getAppMode() !== 'production') {
+              console.info(
+                `ws [${id}] was closed, reconnecting in 2 seconds %c@ ${new Date().toUTCString()}`,
+                'color: #255897;'
```
oskarbauer, line 141: will this color be a common constant?

http-gateway/web/src/components/table/utils.js
```diff
@@ -0,0 +1,12 @@
+// Case insensitive sort function
+export const compareIgnoreCase = (a, b) => {
+  let r1 = a.toLowerCase()
```
oskarbauer, line 3: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare - looks like this is a native function that can take care of the weird things about languages.

http-gateway/web/src/containers/things/websockets.js
```diff
@@ -0,0 +1,56 @@
...
+          status,
+        })
+
+        // Show toast
+        if (notificationsEnabled && status !== thingsStatuses.UNREGISTERED) {
+          const {
+            data: { device: { n: deviceName } = {} } = {},
+          } = await getThingApi(deviceId)
+          const toastMessage =
+            status === thingsStatuses.ONLINE
```
oskarbauer: this variable could be inlined.

oskarbauer (approved):

http-gateway/web/src/common/services/init-services.js
kuby: I might handle it in the future, for now it doesn't need to be, the reconnect logic takes care of it.

http-gateway/web/src/common/services/ws-manager.js
kuby: Its just a color for the console.log and only used in development, nothing special.

http-gateway/web/src/components/table/utils.js
kuby: I will look at it in the future, it needs bigger research since the browser support is not the best. Also, sorting will be asynchronous soon, so there is no need for us to spend that much time on it in JS.

ondrejtomcik (approved):

PR #209, opened 2021-03-06
support auth0 in bundle

ondrejtomcik:
documentation still in progress


bundle/run.sh
```diff
@@ -52,32 +53,81 @@ export LISTEN_FILE_CERT_KEY_NAME="$GRPC_INTERNAL_CERT_KEY_NAME"
...
 export DOMAIN=${FQDN_NGINX_HTTPS}
 if [ "$NGINX_PORT" = "443" ]; then
   export DOMAIN=${FQDN}
 fi
 
+#OAUTH SERVER
+if [ -z "${OAUTH_AUDIENCE}" ]
+then
+export OAUTH_AUDIENCE=test
+export SERVICE_OAUTH_AUDIENCE=${OAUTH_AUDIENCE}
```
jkralik: you can move out line from the if block - it same for both cases if/else

bundle/run.sh
```diff
@@ -52,32 +53,81 @@ export LISTEN_FILE_CERT_KEY_NAME="$GRPC_INTERNAL_CERT_KEY_NAME"
...
+export SDK_OAUTH_AUDIENCE=${FQDN}:${NGINX_PORT}
+else
+export OAUTH_AUDIENCE=${OAUTH_AUDIENCE}
+export SERVICE_OAUTH_AUDIENCE=${OAUTH_AUDIENCE}
```
jkralik: you can move out line from the if block - it same for both cases if/else

bundle/run.sh
```diff
@@ -52,32 +53,81 @@ export LISTEN_FILE_CERT_KEY_NAME="$GRPC_INTERNAL_CERT_KEY_NAME"
...
+export SDK_OAUTH_AUDIENCE=${OAUTH_AUDIENCE}
+fi
+
+if [ -z "${OAUTH_ENDPOINT_AUTH_URL}" ]
+then
+export DEVICE_OAUTH_ENDPOINT_AUTH_URL="https://localhost:${MOCKED_OAUTH_SERVER_PORT}/authorize"
+export OAUTH_ENDPOINT_CODE_URL="https://localhost:${MOCKED_OAUTH_SERVER_PORT}/authorize"
+export SERVICE_CLIENT_CONFIGURATION_AUTHCODEURL="https://${FQDN}:${NGINX_PORT}/authorize?client_id=test"
```
jkralik: pl ident it with 2 whitespaces

jkralik (approved):

PR #215, opened 2021-03-21
[ui] Resources tree view

### LINKED ISSUE #375: Support tree view for resource browsing
Support easier browsing through device resources.

If the device has e.g. 100 resources that have a hierarchy, e.g.
 - `/channels/1`
 - `/channels/1`
 - `/channesl/1/calibrations/32`
 - `/channesl/3/calibrations/1`
It's hard to browse them in a table. Having an expandable tree view would be very nice!

None

### LINKED ISSUE #376: Support device name change
User shall be able to change the device name in the plgd Dashboard.

If a user has more than one device and they have always the same name, it's hard to identify the one she wants to configure. 

None

kuby:
- TreeTable component
- TreeExpander component
- Resources TreeView + switch between normal and tree
- Resource registration bugfix with merging new items which are already present
- Switch style updates
- Tooltip style updates
- Sound updates
- Device name change
- Default device name if not set
- Code refactoring


http-gateway/web/src/common/utils/play-random-fart-sound.js
```diff
@@ -2,11 +2,27 @@ import fart1 from '@/assets/audio/fart1.mp3'
...
 import fart3 from '@/assets/audio/fart3.mp3'
 
+const sound1 = new Audio(fart1)
+const sound2 = new Audio(fart2)
+const sound3 = new Audio(fart3)
+
+export const loadAllFartSounds = () => {
+  const sounds = [sound1, sound2, sound3]
+
+  sounds.forEach(sound => {
```
oskarbauer: Thinking that you should explain why you are doing it this way.

http-gateway/web/src/containers/things/_things-details-title.js
```diff
@@ -0,0 +1,154 @@
...
+      } catch (error) {
+        if (error && isMounted.current) {
+          showErrorToast({
+            title: _(t.thingNameChangeFailed),
+            message: getApiErrorMessage(error),
+          })
+        }
+      }
+    } else {
+      cancel()
```
oskarbauer: maybe something a little more descriptive like "cancelSave"

http-gateway/web/src/containers/things/hooks.js
```diff
@@ -59,7 +59,20 @@ export const useThingDetails = deviceId => {
         let updatedLinks = []
 
         if (event === resourceEventTypes.ADDED) {
-          updatedLinks = data.links.concat(resource)
+          const linkExists =
+            data.links.findIndex(link => link.href === resource.href) !== -1
```
oskarbauer, line 63: maybe use some()

oskarbauer: or maybe just do the map, or reduce.  You seem to be doing the same comparison twice.

http-gateway/web/src/containers/things/utils.js
```diff
@@ -104,6 +115,70 @@ export const updateThingsDataStatus = (data, { deviceId, status }) => {
...
+
+export const createNestedResourceData = data => {
+  // Always construct the objects from scratch
+  let firstSwipe = {}
+  if (data) {
+    data.forEach(item => {
+      addItem(firstSwipe, item, 1)
+    })
+  }
+  const output = deDensisfy(firstSwipe)
```
oskarbauer, line 166: //  Then take the object structure and output the tree scructure

oskarbauer (approved):

http-gateway/web/src/containers/things/hooks.js
kuby: Yes, because once I check if the resource is in the array, if it is, I will have to update it, if it isn't, I need to update it. I cannot achieve this with map nor reduce, since I would be only looping through the resource list I already have.

http-gateway/web/src/common/utils/play-random-fart-sound.js
kuby: Added a comment.

http-gateway/web/src/containers/things/utils.js
kuby: Thanks.

http-gateway/web/src/containers/things/_things-details-title.js
kuby: Changed.

PR #217, opened 2021-03-24
fix get JWK

jkralik:
use http.DefaultTransport.(*http.Transport).Clone()

the issue when it try to retrieves JWK.


der-dano (approved):
good job!

PR #218, opened 2021-03-24
Allow to set owner claim

### LINKED ISSUE #377: Configurable device owner claim
Devices, before they can be used and onboarded to the cloud, are owned by some subject. In the current implementation, this subject is limited to the `sub` claim of the token device uses. Claim used to identify device owner shall be configurable.

In a multitenant system, devices might be owned by the tenant and not by the user who provisions them. Limiting owner to "sub" set the device owner always to that user and not to e.g. tenant.

None, sub claim cannot be modified.

jkralik:


authorization/service/getUserDevices.go
```diff
@@ -67,40 +66,28 @@ func (s *Service) GetUserDevices(request *pb.GetUserDevicesRequest, srv pb.Autho
...
 			return logAndReturnError(status.Errorf(codes.InvalidArgument, "cannot get user devices: invalid userIdsFilter"))
 		}
-		userIdsFilter = []string{userID}
-	}
-
-	deviceIdsFilter := make(map[string]bool)
-	for _, deviceID := range request.GetDeviceIdsFilter() {
-		deviceIdsFilter[deviceID] = true
+		if owner == serviceOwner {
+			return sendUserDevices(request, srv, tx.RetrieveAll)
```
ondrejtomcik, line 77: RA or RD, services using service token can fill userIdsFilter or deviceIdsFilter. Why are you ignoring values from there ?

bundle/Dockerfile
```diff
@@ -114,8 +113,8 @@ ENV DEVICE_PROVIDER_TYPE=plgd
 ENV DEVICE_OAUTH_SCOPES=profile,openid,offline_access
 ENV SDK_OAUTH_RESPONSE_MODE=form_post
 ENV SDK_OAUTH_SCOPES=openid
+ENV OWNER_CLAIM="sub"
 ENV SERVICE_CLIENT_CONFIGURATION_CLOUDAUTHORIZATIONPROVIDER="plgd"
-ENV SERVICE_CLIENT_CONFIGURATION_JWTCLAIMOWNERID="sub"
```
ondrejtomcik, line 118: `SERVICE_CLIENT_CONFIGURATION_JWTCLAIMOWNERID` doesn't have to match `OWNER_CLAIM`. You have a use-case when service uses sub - user id for a user filter, but authorization service - custom implementation like Identity Server uses different claim, e.g. Tenant.

pkg/net/coap/auth.go
```diff
@@ -0,0 +1,48 @@
+package coap
```
ondrejtomcik: I am in favor of having a repository structure that structures things based on their primary functionality of domain in the business logic. That means, and it's for 95% possible, place these "utils" to the package where they from business perspective belong to.
In this case, coap-gateway.

pkg/net/grpc/auth.go
```diff
@@ -0,0 +1,152 @@
+package grpc
```
ondrejtomcik, line 1: grpc-gateway

pkg/net/http/auth.go
```diff
@@ -0,0 +1,72 @@
+package http
```
ondrejtomcik, line 1: http gateway

pkg/security/jwt/claims.go
```diff
@@ -0,0 +1,56 @@
+package jwt
```
ondrejtomcik, line 1: authorization-server

pkg/security/oauth/manager/config.go
```diff
@@ -0,0 +1,38 @@
+package manager
```
ondrejtomcik, line 1: authorization-server

authorization/service/getUserDevices.go
jkralik: userIdsFilter is handled by https://github.com/plgd-dev/cloud/pull/218/files#diff-c6adc03507ee9569cc1a9e1ed0004c4b456aa337c574fcdf9e8f5bce78c94f77R64

devicesIdsFilter is handled by https://github.com/plgd-dev/cloud/pull/218/files#diff-c6adc03507ee9569cc1a9e1ed0004c4b456aa337c574fcdf9e8f5bce78c94f77R33

pkg/net/coap/auth.go
jkralik: moved to coap-gw

ondrejtomcik (approved):

PR #225, opened 2021-03-30
add hostname SAN during certificate generation

makothap:
Request to add the hostname as an alternative subject name in Certificate generation. In our scenario, we're launching the plgd/bundle container with a custom hostname (say, as "msc"), and we would like other apps in the same docker network to refer bundle/gateway container endpoints using https://ctr-hostname:443/<context>. Thanks!


ondrejtomcik:
@makothap this is already outdated as we modified the `run.sh `. As we will soon helm chart, I think this won't be relevant for you. Jozef anyway found out that this would break something. If it's not okay to close it and use than helm chart, please reopen with updated MR.

PR #229, opened 2021-04-05
Config refactoring yaml envconfig

Trishia:
This is a branch where both yaml and envconfig were applied and also merged with latest commit (https://github.com/plgd-dev/cloud/tree/b5ed2cac7b7f777608006baad99f3f6cce1e66a7 on Mar 27) of v2 branch. If you need to continue config-refactoring, you'd better use this. This branch was merged with --squash option from my private developing branch so you can not see any history commit after config-refactoring branch.


Trishia:
This pull requesting of config-refactoring-yaml-envconfig branch will be closed due to other branches under processing config-refactoring-{each services (i.e config-refactoring-resource-directory)}.

PR #242, opened 2021-04-15
configure resource-directory service via yaml file 

jkralik:


resource-directory/config.yaml
```diff
@@ -0,0 +1,94 @@
...
+        maxConnsPerHost: 32
+        maxIdleConnsPerHost: 16
+        idleConnTimeout: 30s
+        timeout: 10s
+        tls:
+          caPool: "/secrets/public/rootca.crt"
+          keyFile: "/secrets/private/cert.key"
+          certFile: "/secrets/public/cert.crt"
+          useSystemCAPool: false
+exposedCloudConfiguration:
```
ondrejtomcik: exposed -> public

ondrejtomcik (approved):

resource-directory/config.yaml
jkralik: Fixed as `publicConfiguration`.

PR #252, opened 2021-05-03
configure grpc-gateway service via yaml file

jkralik:


grpc-gateway/config.yaml
```diff
@@ -0,0 +1,60 @@
...
+    address: "0.0.0.0:9100"
+    tls:
+      caPool: "/secrets/public/rootca.crt"
+      keyFile: "/secrets/private/cert.key"
+      certFile: "/secrets/public/cert.crt"
+      clientCertificateRequired: true
+    authorization:
+      authority: ""
+      audience: ""
+      ownerClaim: sub
```
ondrejtomcik: Forgot, why do we have it here as it is relevant only for authorization-server which parses the token?

resource-directory/config.yaml
```diff
@@ -25,6 +25,8 @@ apis:
           useSystemCAPool: false
 clients:
   eventBus:
+    # number of routines to process events in projection
+    goPoolSize: 16
```
ondrejtomcik, line 28: Why you added it only to RD and not RA ?
Please update also documentation Trishia already did for RD. It's in v2.

resource-directory/config.yaml
```diff
@@ -35,8 +37,6 @@ clients:
   eventStore:
     # expiration time of cached resource in projection
     cacheExpiration: 1m
-    # number of routines to process events in projection
-    goPoolSize: 16
```
ondrejtomcik, line 39: And this remove from doc.

ondrejtomcik (changes requested):

grpc-gateway/config.yaml
jkralik: removed

resource-directory/config.yaml
jkralik: updated doc - new evenstore doesn't use goPool.

resource-directory/config.yaml
jkralik: updated

authorization/config.yaml
```diff
@@ -32,6 +31,7 @@ apis:
       clientCertificateRequired: true
 oauthClients:
   device:
+    ownerClaim: sub
```
ondrejtomcik: It doesn't belong to oauthClients.device. When we move this configuration to coapgateway, you will have to add it to other place in this authorization-server config - again breaking change. And, coap-gateway doesn't need it anyway. It's the general configuration of the authorization server.

CoAP gateway just sends "register device" request to authorization server with token, it won't care about jwt claims.

ondrejtomcik (changes requested):

authorization/config.yaml
ondrejtomcik: As discussed, will be moved to storage group.

ondrejtomcik (approved):

PR #255, opened 2021-05-06
configure test/oauth-server service via yaml file

jkralik:
Co-authored-by: Trishia Yun <trishia1298@imsoftlab.com>

### DIFF
--- a/test/oauth-server/config.yaml
+++ b/test/oauth-server/config.yaml
@@ -1,0 +1,14 @@
+log:
+  debug: false
+apis:
+  http:
+    address: "0.0.0.0:9100"
+    tls:
+      caPool: "/secrets/public/rootca.crt"
+      keyFile: "/secrets/private/cert.key"
+      certFile: "/secrets/public/cert.crt"
+      clientCertificateRequired: true
+oauthSigner:
+  idTokenKeyFile: "/secrets/private/idToken.key"
+  accessTokenKeyFile: "/secrets/private/accessToken.key"
+  domain: ""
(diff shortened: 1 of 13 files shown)


test/oauth-server/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+apis:
+  http:
+    address: "0.0.0.0:9100"
+    tls:
+      caPool: "/secrets/public/rootca.crt"
+      keyFile: "/secrets/private/cert.key"
+      certFile: "/secrets/public/cert.crt"
+      clientCertificateRequired: true
+oauthSigner:
+  idTokenKeyFile: "/secrets/private/idToken.key"
```
ondrejtomcik, line 12: Why are you signing tokens with different certs?

test/oauth-server/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  accessTokenKeyFile: "/secrets/private/accessToken.key"
+  domain: ""
```
ondrejtomcik, line 14: What is this domain used for?

test/oauth-server/config.yaml
jkralik: becase idToken must by sigend by RSA (ui library doesn't supports ECDSA)
and accessToken we signed by ECDSA (is better and smaller)

test/oauth-server/config.yaml
jkralik: It used as:

- issuer for idToken and accessToken
- audience for accessToken

test/oauth-server/config.yaml
ondrejtomcik: Why exactly is ECDSA better if the oauth-server is used only in tests and bundle (example for development)? Why we make deployment and configuration more complex?

test/oauth-server/config.yaml
ondrejtomcik: @Trishia please document what exactly is it used for.

ondrejtomcik (approved):

PR #268, opened 2021-05-14
coap-gateway: update pending commands via eventbus

jkralik:

PR #269, opened 2021-05-14
coap-gateway: allow deregister without token for signed in device

jkralik:


coap-gateway/service/signUp.go
```diff
@@ -153,6 +153,9 @@ func signOffHandler(req *mux.Message, client *Client) {
 	if deviceID == "" {
 		deviceID = authCurrentCtx.GetDeviceID()
 	}
+	if accessToken == "" {
```
ondrejtomcik: Why are you hiding SignOff in signUp.go ?

coap-gateway/service/signUp.go
```diff
@@ -153,6 +153,9 @@ func signOffHandler(req *mux.Message, client *Client) {
...
+		accessToken = authCurrentCtx.GetAccessToken()
+	}
 
 	err := validateSignOff(deviceID, accessToken)
```
ondrejtomcik: Inside of validateSignOff is error message bit weird. If access token is empty, error message is "invalid  ". What is invalid?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.15-alpine AS cloud-build
+FROM golang:1.16-alpine AS cloud-build
```
ondrejtomcik: The build is now failing. Is the change from Daniel in plgd branch? Are we will using plgd branch or are we on master? If we are on master, let's wait until it's get there.

jkralik: Sometimes onboard/disown via sdk fails - it will be fixed by upgrade sdk in PR https://github.com/plgd-dev/cloud/pull/268

coap-gateway/service/signUp.go
jkralik: moved to signOff.go

coap-gateway/service/signUp.go
jkralik: fixed

ondrejtomcik (approved):

PR #280, opened 2021-06-01
oauth-manager: fix evaluation of should refresh function

jkralik:


pkg/security/oauth/manager/manager.go
```diff
@@ -15,16 +15,16 @@ import (
...
-	token             *oauth2.Token
-	httpClient        *http.Client
-	tokenErr          error
-	doneWg            sync.WaitGroup
-	done              chan struct{}
+	mutex            sync.Mutex
+	config           clientcredentials.Config
+	requestTimeout   time.Duration
+	tickFrequency    time.Duration
+	nextRefreshToken time.Time
```
der-dano: Everything's OK, I would just rename this field - `nextRefreshToken` is hard to understand without context...

jkralik: fixed - renamed `nextRefreshToken` to `nextTokenRenewalTime`

der-dano (approved):

PR #281, opened 2021-06-02
use device metadata instead of cloud status resource

jkralik:


jkralik:
It is already in PR https://github.com/plgd-dev/cloud/pull/287

PR #289, opened 2021-06-15
Resource-aggregate: skip duplicated events

Danielius1922:


resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -172,8 +173,28 @@ func (e *ResourceStateSnapshotTaken) ValidateSequence(eventMetadata *EventMetada
...
+	}
+
+	if current.Content.CoapContentFormat != changed.Content.CoapContentFormat ||
+		current.Content.ContentType != changed.Content.ContentType ||
+		!bytes.Equal(current.Content.Data, changed.Content.Data) {
+		return false
+	}
+
+	if current.AuditContext.UserId != changed.AuditContext.UserId ||
+		current.AuditContext.CorrelationId != changed.AuditContext.CorrelationId {
```
jkralik: pls remove comparison of `CorrelationID` it can be different

resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -172,8 +173,28 @@ func (e *ResourceStateSnapshotTaken) ValidateSequence(eventMetadata *EventMetada
...
+		return false
+	}
+
+	return true
+}
+
 func (e *ResourceStateSnapshotTaken) HandleEventResourceChanged(ctx context.Context, changed *ResourceChanged) (bool, error) {
-	if e.ValidateSequence(changed.GetEventMetadata()) {
+	if e.ValidateSequence(changed.GetEventMetadata()) &&
+		(e.LatestResourceChange != nil && !Equal(e.LatestResourceChange, changed)) {
```
jkralik: and what `if e.LatestResourceChange == nil` ?

jkralik (commented):
pls fix comments. thx

resource-aggregate/events/resourceStateSnapshotTaken.go
Danielius1922: yes, this is incorrect, if `LatestResourceChange == nil` then we shouldn't skip the event, we just don't have the cache value set

jkralik (approved):

resource-aggregate/events/resourceLinksSnapshotTaken.go
```diff
@@ -45,7 +45,35 @@ func (e *ResourceLinksSnapshotTaken) IsSnapshot() bool {
...
 }
 
-func (e *ResourceLinksSnapshotTaken) HandleEventResourceLinksPublished(ctx context.Context, pub *ResourceLinksPublished) error {
+// Check if the snapshot of published resources matches exactly the new published data
+func (e *ResourceLinksSnapshotTaken) MatchesLinksPublished(published *ResourceLinksPublished) bool {
+	if published == nil {
+		return false
+	}
+
+	if len(e.Resources) != len(published.Resources) {
```
ondrejtomcik: ResourceLinkPublish operation can be by the devices split into multiple publish messages. This condition will automatically consider the received subset as not matching one, even though 3 published resources are already stored and were not changed. Duplicate events are published.

resource-aggregate/events/resourceLinksSnapshotTaken.go
```diff
@@ -45,7 +45,35 @@ func (e *ResourceLinksSnapshotTaken) IsSnapshot() bool {
...
+		return false
+	}
+
+	for _, resPub := range published.GetResources() {
+		resSnap, ok := e.Resources[resPub.Href]
+		if !ok {
+			return false
```
ondrejtomcik: 10 resources {a,b,c...j} are already published. 5 resources are being published by the device after one hour because user was playing with the device when it was disconnected from the cloud. These 5 resources which are contained in the publish request are {a,b,x,y,z}.
What happens is (assuming `if len(e.Resources) != len(published.Resources)` is removed) based on the order, either user is notified about all 5 resources or not at all.

This function shouldn't return true/false, but it should return diff of resources which are not yet published - which are not yet part of the `e`.

The same is done in the `HandleEventResourceLinksUnpublished`, which notifies user only about those unpublished resources, which were really unpublished.

ondrejtomcik (changes requested):

resource-aggregate/events/resourceLinksSnapshotTaken.go
Danielius1922: Reworked to return slice of resources not yet cached and updated resources

resource-aggregate/events/resourceLinksSnapshotTaken.go
```diff
@@ -146,9 +166,9 @@ func (e *ResourceLinksSnapshotTaken) HandleCommand(ctx context.Context, cmd aggr
 			AuditContext:  ac,
 			EventMetadata: em,
 		}
-		err := e.HandleEventResourceLinksPublished(ctx, &rlp)
-		if err != nil {
-			return nil, err
+		ok := e.HandleEventResourceLinksPublished(ctx, &rlp)
```
ondrejtomcik: &rlp is sent to the eventStore and eventBus. As far as I understand your code, it will every time send and event which contains also resourceLinks which were already published. Shouldn't the final event contain only resource links which were published within one command execution?

That means, `HandleEventResourceLinksPublished` not only verifies and appends the dif to snapshot, but also returns resources which were published. Currently it returns true so you will send an event which contains resources which were already published. 

In case of unpublish, unpublished events are deleted from the snapshot struct and items which were deleted are returned to the eventstore/eventbus.
In my opinion same should happen for the publish, what was added is appended (this is done), and only published resources are returned to the eventstore/eventbus.
Please confirm @jkralik

Danielius1922: You are correct, I didn't commit all the changes I did by mistake.

ondrejtomcik (approved):

resource-aggregate/events/resourceLinksSnapshotTaken.go
```diff
@@ -146,10 +168,11 @@ func (e *ResourceLinksSnapshotTaken) HandleCommand(ctx context.Context, cmd aggr
 			AuditContext:  ac,
 			EventMetadata: em,
 		}
-		err := e.HandleEventResourceLinksPublished(ctx, &rlp)
+		published, err := e.HandleEventResourceLinksPublished(ctx, &rlp)
 		if err != nil {
 			return nil, err
 		}
+		rlp.Resources = published
 		return []eventstore.Event{&rlp}, nil
```
jkralik, line 179: So you always create event even published array is empty .

jkralik (changes requested):
Pls fix handle command it produce event only when published is not empty.

jkralik (approved):

PR #290, opened 2021-06-16
coap-gateway/resourceDiscovery: fix time to live checks

Danielius1922:
Use value of -1 instead of 0 to signal that TimeToLive or
TimeToLiveLegacy are not set. Value of 0 is used by
iotivity-lite to signal that TimeToLive is unlimited.

### DIFF
--- a/coap-gateway/service/resourceDirectory.go
+++ b/coap-gateway/service/resourceDirectory.go
@@ -23,9 +23,16 @@ type wkRd struct {
 	TimeToLiveLegacy int                  `json:"lt"`
 }
 
+func makeWkRd() wkRd {
+	return wkRd{
+		TimeToLive:       -1,
+		TimeToLiveLegacy: -1,
+	}
+}
+
 func fixTTL(w wkRd) wkRd {
 	// set time to live properly
-	if w.TimeToLive <= 0 {
+	if w.TimeToLive < 0 {
 		w.TimeToLive = w.TimeToLiveLegacy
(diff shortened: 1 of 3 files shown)


coap-gateway/service/resourceDirectory.go
```diff
@@ -65,7 +72,7 @@ func resourceDirectoryPublishHandler(req *mux.Message, client *Client) {
 		return
 	}
 
-	var w wkRd
+	w := makeWkRd()
```
ondrejtomcik, line 75: Why to define a new func which is used just once and not directly assign
```
w := &wkRd{
		TimeToLive:       -1,
		TimeToLiveLegacy: -1,
	}
```

jkralik (approved):

coap-gateway/service/resourceDirectory.go
Danielius1922: `TimeToLive` and `TimeToLiveLegacy` should always be initialized to -1, I have no way of enforcing it since Go doesn't do constructors, so this pseudo-constructor should always be used when wkRd is being initialized. Though this is not ideal since a dev can not notice the pseudo-constructor and call `&wkRd{...}` directly and not set the `-1` values. We could change `TimeToLive` and `TimeToLiveLegacy` to pointers or translate 0 to some other value that would signify unlimited TTL if this seems too fragile.

ondrejtomcik (approved):

PR #296, opened 2021-06-25
resource-aggregate: fix tests

Danielius1922:
Duplicate publish and unpublish events are skipped and return
nil instead of array with empty value.


jkralik (commented):
It seems that there is a bug in Makefile.

Danielius1922:
Cherry-picked to the http-gateway-provides-metadata branch

PR #300, opened 2021-06-30
resource-aggregate: allow publishing to jetstream

### LINKED ISSUE #149: Jetstream as an EventBus
Some users are interested in persisting streams of events. JetStream which is built on top of NATS can be used to achieve this in a way it won't affect the plgd cloud architecture.

Providing persisted streams of events what is interested for some clients subscribed directly to NATS/JetStream.

None

jkralik:


bundle/Dockerfile
```diff
@@ -82,6 +87,8 @@ COPY --from=build /go/bin/certificate-authority /usr/local/bin/certificate-autho
 COPY --from=cloud-build /go/src/github.com/plgd-dev/cloud/certificate-authority/config.yaml /configs/certificate-authority.yaml
 COPY --from=build /go/bin/certificate-generator /usr/local/bin/certificate-generator
 COPY --from=build /go/bin/nats-server /usr/local/bin/nats-server
+COPY --from=build /go/bin/nats /usr/local/bin/nats
+COPY --from=cloud-build /go/src/github.com/plgd-dev/cloud/bundle/jetstream.json /configs/jetstream.json
```
ondrejtomcik, line 91: plgd service configuration is exposed using environment variables. In the case of the jetstream, it's only this file. Users might be interested in updating it by mounting a volume, but if they do it, they remove all service configuration files. Add it one level deeper, /configs/jetstream/jetstram.json so in case of users want to change it, they will map just this volume containing the jetstream configuration.

bundle/jetstream.json
```diff
@@ -0,0 +1,17 @@
+{
```
ondrejtomcik, line 1: Please include `JetStream part of the plgd bundle`before this one: https://github.com/plgd-dev/plgd.dev/blob/main/content/en/features/jetstream.md#deploy-jetstream-controller-for-k8s

and document how to enable it and how to change the configuration and how to configuration looks like (one from run.sh - limits.

docs/guide/deployment/resource-aggregate.md
```diff
@@ -154,6 +154,7 @@ Plgd cloud uses NATS messaging system as a event bus.
 | Property | Type | Description | Default |
 | ---------- | -------- | -------------- | ------- |
 | `clients.eventBus.nats.url` | string | `URL to nats messaging system.` | `"nats://localhost:4222"` |
+| `clients.eventBus.nats.jetstream `| bool | `If true, events will be published to jetstream.` | `false` |
```
ondrejtomcik, line 157: Does the connection string stay the same (just pointing to jetstream? is there any difference?)? It's still eventbus.nats which needs to be configured?

resource-aggregate/cqrs/eventbus/nats/subscriber/subscriber_test.go
```diff
@@ -18,10 +18,10 @@ import (
 )
 
 func TestSubscriber(t *testing.T) {
-	topics := []string{"test_subscriber_topic0" + uuid.Must(uuid.NewV4()).String(), "test_subscriber_topic1" + uuid.Must(uuid.NewV4()).String()}
+	publishTopics := []string{"test.subscriber.topic0." + uuid.Must(uuid.NewV4()).String(), "test.subscriber.topic1." + uuid.Must(uuid.NewV4()).String()}
+	subscriberTopics := []string{ /*"test.subscriber.topic0.>",*/ "test.subscriber.*.>", "test.subscriber.*.>"}
```
ondrejtomcik: shouldn't be also this one `test.subscriber.topic0.>` uncommented?

ondrejtomcik (changes requested):

bundle/Dockerfile
jkralik: Docker allows you to mount file instead of directory

jkralik: https://stackoverflow.com/questions/42248198/how-to-mount-a-single-file-in-a-volume

docs/guide/deployment/resource-aggregate.md
jkralik: Yes. It same for jetstream.

ondrejtomcik (approved):

PR #301, opened 2021-06-30
Add LatestTimestamp field to eventstore documents in db

Danielius1922:


resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
```diff
@@ -252,7 +270,24 @@ func getLatestSnapshotVersion(events []eventstore.Event) (uint64, error) {
 		}
 	}
 	return latestSnapshotVersion, err
+}
+
+func getLatestTimestamp(events []eventstore.Event) (int64, error) {
+	// TODO: can we guarantee that the last event also has the latest timestamp? If yes,
+	// then this algorithm is not necessary and we only need to take the timestamp
+	// from the last event
```
Danielius1922: @jkralik, can we guarantee this (that the last event in the events array has the latest timestamp)? If so I can just use `events[len(events)-1].Timestamp().UnixNano()` and remove this whole method.

jkralik (approved):

PR #332, opened 2021-07-06
resource-aggregate: add basic GetEvents query

Danielius1922:
GetEvents query return list of events for given device and/or
resource that occured later than given timestamp. All parameters
(device, resource or timestamp) are optional. If some are not
set then no filtering based on that field will be done. If none
are set then all events for all devices are returned.


resource-aggregate/cqrs/eventstore/mongodb/eventstore_test.go
```diff
@@ -40,4 +40,8 @@ func TestEventStore(t *testing.T) {
 
 	t.Log("event store with default namespace")
 	test.AcceptanceTest(t, ctx, store)
+
+	t.Log("clearing db")
+	store.Clear(ctx)
```
jkralik: This cleanup store. Cloud you create a separate test?

resource-aggregate/cqrs/eventstore/mongodb/getevents.go
```diff
@@ -0,0 +1,99 @@
...
+
+func getNormalizedGetEventsQuery(queries []eventstore.GetEventsQuery) map[string][]eventstore.GetEventsQuery {
+	normalizedQuery := make(map[string][]eventstore.GetEventsQuery)
+	for _, query := range queries {
+		v, ok := normalizedQuery[query.GroupID]
+		if !ok {
+			v = make([]eventstore.GetEventsQuery, 0, 1)
+		}
+		v = append(v, query)
+		normalizedQuery[query.GroupID] = v
```
jkralik: need to be properly implemented

Danielius1922: Do you mean some changes to this method? Because currently this just splits the mongodb query based on DeviceId (GroupId).

resource-aggregate/cqrs/eventstore/test/acceptance_testing.go
```diff
@@ -135,6 +184,84 @@ const groupID1 = "deviceId1"
...
+	require.True(t, saveEh.Equals(groupID3Events))
+
+	timestamp = timestamp2 - 1
+	t.Logf("get aggregateid (%v, %v) events with timestamp > %v", aggregateID3, aggregateID4, timestamp)
+	saveEh = NewMockEventHandler()
+	store.GetEvents(ctx, []eventstore.GetEventsQuery{{AggregateID: aggregateID3}, {AggregateID: aggregateID4}}, timestamp, saveEh)
+	events = groupID2AggID3Events
+	events = append(events, groupID3Events...)
+	require.NoError(t, err)
+	require.True(t, saveEh.Equals(events))
```
jkralik, line 152: pls add there test for usecase: GroupID + timestamp

resource-aggregate/cqrs/eventstore/mongodb/getevents.go
jkralik: When you have Queries: one with AggregateId and GroupId and the second query with the same GroupId, then you absorb the first one.

resource-aggregate/cqrs/eventstore/mongodb/getevents.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+
+	return projection
+}
+
+func getEventsQueriesToMongoQuery(groupID string, queries []eventstore.GetEventsQuery, timestamp int64) (interface{}, *options.FindOptions) {
+	filter := getEventsFilter(groupID, queries, timestamp)
+
+	opts := options.Find()
```
jkralik, line 100: Why you don't use index-es ?
If you add indexes could you compare duration time of test `https://github.com/plgd-dev/cloud/blob/v2/resource-aggregate/cqrs/aggregate/aggregateParralel_test.go#L50` with and without indexes ?

resource-aggregate/cqrs/eventstore/mongodb/getevents_test.go
```diff
@@ -0,0 +1,28 @@
...
+func Test_getNormalizedGetEventsFilter(t *testing.T) {
+	type args struct {
+		queries []eventstore.GetEventsQuery
+	}
+	tests := []struct {
+		name string
+		args args
+		want deviceIdFilter
+	}{
+		// TODO: Add test cases.
```
jkralik: Testcase

resource-directory/service/getEvents.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	var queries []eventstore.GetEventsQuery
+	if len(req.DeviceIdFilter) == 0 && len(req.ResourceIdFilter) == 0 {
+		queries = getUserDeviceQueries(mapUserDeviceIds)
+	} else {
+		queries = getDeviceQueries(req.DeviceIdFilter, mapUserDeviceIds)
+		queries = append(queries, getResourceQueries(req.ResourceIdFilter, mapUserDeviceIds)...)
+	}
+	// TODO indexes
```
jkralik: this TODO belows to mongodb eventstore.

resource-aggregate/cqrs/eventstore/test/acceptance_testing.go
Danielius1922: Added

resource-aggregate/cqrs/eventstore/mongodb/getevents.go
Danielius1922: Is using `hint` necessary though? Mongodb uses indexes automatically, using hint (https://docs.mongodb.com/manual/reference/operator/meta/hint/) forces the query optimizer to use the selected index. Is the query optimizer slow or why is `hint` necessary?

resource-aggregate/cqrs/eventstore/mongodb/getevents_test.go
Danielius1922: Added

resource-aggregate/cqrs/eventstore/mongodb/getevents.go
jkralik: Yes. From my experience if I don't provide hint it uses full scan instead of indexes and find query was very slow.

You can try: - pls create test where you store lot's of events (200000 events) and query half of them via getEvents call and measure the duration. (eg https://github.com/plgd-dev/cloud/blob/v2/resource-aggregate/cqrs/aggregate/aggregateParralel_test.go#L50)

jkralik (approved):

PR #337, opened 2021-07-10
golangci-sci fixes in bundle and pkg

Danielius1922:


bundle/client/coap/main.go
```diff
@@ -235,7 +241,10 @@ func main() {
 		if err != nil {
 			log.Fatalf("cannot observe resource: %v", err)
 		}
-		defer obs.Cancel(context.Background())
+		defer func() {
+			err := obs.Cancel(context.Background())
+			fmt.Printf("failed to cancel context: %v", err)
```
jkralik: failed to cancel **observation**

Danielius1922: line 245?

Danielius1922: Fixed

jkralik (approved):

PR #338, opened 2021-07-10
Fixes for proto/generate

Danielius1922:


grpc-gateway/Makefile
```diff
@@ -23,20 +25,23 @@ push: build-servicecontainer
 	docker push plgd/$(SERVICE_NAME):$(VERSION_TAG)
 	docker push plgd/$(SERVICE_NAME):$(LATEST_TAG)
 
+GOOGLEAPIS_MODULE_PATH := $(shell find $(GOPATH)/pkg/mod/github.com/googleapis -mindepth 1 -maxdepth 1 -iname 'googleapi*' -print -quit)
```
jkralik, line 28: Why you not using another steps than GRPC_GATEWAY_..

Danielius1922: `go list` works only for Go modules or packages, googleapis is neither, it is just a git repo, so Go cannot track it. I guess checkouting out googleapis as a submodule would be cleaner. 

In the [grpc-gateway guide](https://github.com/grpc-ecosystem/grpc-gateway) they suggest manually cloning and copying `google/api/annotations.proto`. Or they suggest using [buf](https://github.com/bufbuild/buf) instead of `protoc`.

jkralik (approved):

PR #342, opened 2021-07-14
Fix issues in coap-gateway, grpc-gateway and http-gateway

Danielius1922:


coap-gateway/service/clientRetrieveHandler_test.go
```diff
@@ -78,6 +78,7 @@ func Test_clientRetrieveHandler(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), TestExchangeTimeout)
 			defer cancel()
 			req, err := tcp.NewGetRequest(ctx, tt.args.path)
+			assert.NoError(t, err)
```
jkralik: change to `require` - if err occurs req is nil so will be panic when you use it after

ondrejtomcik (approved):

jkralik (approved):

PR #343, opened 2021-07-17
[UI] - WebSockets and REST API rework adaptation

### LINKED ISSUE #364: Configurable command timeout
`UPDATE`, `CREATE`, `DELETE` and `RETRIEVE_FROM_DEVICE` commands are registered as pending commands and will stay there forever if the device is offline. We would like to be able to configure the command time to live, so it expires at some point and won't stay as a pending one forever.

Update command we send to the device shall expire if the device is offline for more than 1 minute. 

Cancel the command on our own after one minute, but there is currently no API to cancel the pending command.

kuby:
- New EDS based WebSocket client
- New fetch method capable of processing data streams
- Implemented fetch cancellation after certain period of time
- Implemented shadowSynchronization controls and states
- Updated dependencies


http-gateway/web/src/common/services/stream-api.js
```diff
@@ -0,0 +1,61 @@
...
+          function push() {
+            // "done" is a Boolean and value a "Uint8Array"
+            reader.read().then(({ done, value }) => {
+              // If there is no more data to read
+              if (done) {
+                controller.close()
+                return
+              }
+              // Get the data and send it to the browser via the controller
+              controller.enqueue(value)
```
jkralik, line 41: Is it possible parse json here  instead of added value ? Otherwise you need to parse it as whole document. If we changed json encoder `\n\n` can be problem in future.

jkralik (approved):

http-gateway/web/src/common/services/stream-api.js
```diff
@@ -0,0 +1,60 @@
...
+    .then(stream => {
+      // Respond with our stream
+      return new Response(stream, {
+        headers: { 'Content-Type': 'text/html' },
+      }).text()
+    })
+    .then(result => {
+      // Parse the result to an array of objects
+      return { data: parseStreamedData(result) }
+    })
```
oskarbauer, line 59: should you have a catch in case something goes wrong?

http-gateway/web/src/common/services/websocket-event-client.js
```diff
@@ -0,0 +1,274 @@
...
+      }
+
+      // Only enable the event listener after a certain time period, to exclude some initial event messages
+      setTimeout(() => {
+        if (this.events[correlationId]) {
+          this.events[correlationId].listenerEnabled = true
+        }
+      }, this.delayListenersMs)
+    } else {
+      console.log('already subscribed')
```
oskarbauer: maybe a more verbose method would be helpful

http-gateway/web/src/common/services/websocket-event-client.js
```diff
@@ -0,0 +1,274 @@
...
+  onClose = () => {}
+
+  onError = () => {}
+}
+
+export const WebSocketEventClient = new _WebSocketEventClient(
+  '/api/v1/ws/events'
+)
+
+/*
```
oskarbauer, line 232: remove the commented out code?

http-gateway/web/src/common/utils/parse-streamed-data.js
```diff
@@ -0,0 +1,11 @@
+// Return an array of objects which came in as a text stream of results
+export const parseStreamedData = stream =>
+  stream
+    ? stream
+        .trim()
+        .split('\n\n')
```
oskarbauer: Messages are separated by\n\n?

http-gateway/web/src/containers/things/_things-resources-list.js
```diff
@@ -51,7 +51,7 @@ export const ThingsResourcesList = ({
       },
       {
         Header: _(t.types),
-        accessor: 'rt',
+        accessor: 'resourceTypes',
```
oskarbauer, line 54: resourceTypes is much better than rt

http-gateway/web/src/containers/things/things-i18n.js
```diff
@@ -155,13 +163,22 @@ export const messages = defineMessages({
...
   },
   resourceDeleteError: {
     id: 'things.resourceDeleteError',
     defaultMessage: 'Failed to delete a resource',
     description: 'Title of the toast message on resource delete error.',
   },
+  shadowSynchronizationError: {
+    id: 'things.shadowSynchronizationError',
+    defaultMessage: 'Failed to set shadow synchronization',
+    description: 'Title of the toast message on shadow synchronization set error.',
```
oskarbauer: Will this be set later?

oskarbauer (approved):
minor stuff to change

http-gateway/web/src/common/services/stream-api.js
kuby: I added a try/catch into the parseStreamedData method, so in case something goes wrong, it return an empty array.

http-gateway/web/src/common/services/websocket-event-client.js
kuby: Added more verbose message, plus it only shows up in dev mode.

http-gateway/web/src/common/services/websocket-event-client.js
kuby: It's an example usage, I made it more clear with another comment.

http-gateway/web/src/common/utils/parse-streamed-data.js
kuby: They are, it's a long story, but in short, the stream sends you binary data, decoded to text, which then must be converted to JSON, that's why the `\n\n`, cause its separated with it.

http-gateway/web/src/containers/things/things-i18n.js
kuby: There are two types of warning titles: `shadowSynchronizationError` and `shadowSynchronizationWasSetOffline`. The one named "offline" is the one which will be set later, the first one is shown, when something went wrong while the device was online.

PR #352, opened 2021-07-28
authorization: add error checks

Danielius1922:


test/oauth-server/service/token.go
```diff
@@ -23,6 +23,8 @@ const (
 	TokenNicknameKey = "nickname"
 	TokenNameKey     = "name"
 	TokenPictureKey  = "picture"
+	TokenEmailKey    = "email"
+	TokenClientIdKey = "client_id"
```
jkralik: where they are used ?

jkralik (approved):

PR #353, opened 2021-07-28
coap-gw: SignIn using OAuth2

### LINKED ISSUE #380: Support JWT tokens on devices

None

Danielius1922:


coap-gateway/service/refreshToken.go
```diff
@@ -42,7 +42,7 @@ func validUntilToExpiresIn(v int64) int64 {
 	if v == 0 {
 		return -1
 	}
-	validUntil := time.Unix(0, v)
+	validUntil := time.Unix(v, 0)
```
jkralik: why you changed the order of args?

coap-gateway/service/signIn.go
```diff
@@ -94,40 +97,93 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
+	v, ok := profile["sub"]
+	if !ok {
+		return 0, fmt.Errorf("invalid userinfo response: %v", "ownerClaim not set")
+	}
+	userID, ok := v.(string)
+	if !ok {
+		return 0, fmt.Errorf("invalid userinfo response: %v", "invalid ownerClaim value")
+	}
+	if userID != signIn.UserID {
+		return 0, fmt.Errorf("invalid userID: %v", signIn.UserID)
```
jkralik: use `owenrClaim` instead of userID - generally, the proper name is the owner.

coap-gateway/service/signIn.go
```diff
@@ -94,40 +97,93 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
+		}
+	}()
+
+	var profile map[string]interface{}
+	if err = json.NewDecoder(resp.Body).Decode(&profile); err != nil {
+		return 0, fmt.Errorf("failed to decode userinfo request body: %v", err)
 	}
 
-	expiresIn := validUntilToExpiresIn(resp.GetValidUntil())
+	v, ok := profile["sub"]
```
jkralik: use `ownerClaim` instead of sub

coap-gateway/service/signIn.go
```diff
@@ -94,40 +97,93 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
+	v, ok = profile["exp"]
+	if !ok {
+		return 0, fmt.Errorf("invalid userinfo response: %v", "exp not set")
+	}
+
+	exp, ok := v.(float64) // all integers are float64 in json
+	if !ok {
+		return 0, fmt.Errorf("invalid userinfo response: %v", "invalid exp value")
+	}
+	return int64(exp), nil
```
jkralik: return `time.Time` instead of `int64`

coap-gateway/service/signIn.go
```diff
@@ -94,40 +97,93 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
 		return
 	}
-	out, err := encode(coapResp)
+
+	expiresIn := validUntilToExpiresIn(validUntil)
+
+	accept, out, err := getContent(expiresIn, req.Options)
 	if err != nil {
-		client.logAndWriteErrorResponse(fmt.Errorf("cannot handle sign in: %w", err), coapCodes.InternalServerError, req.Token)
+		client.logAndWriteErrorResponse(err, coapCodes.InternalServerError, req.Token)
```
jkralik: why you drop `fmt.Errorf("cannot handle sign in: %w", err)`

coap-gateway/service/signIn.go
```diff
@@ -94,40 +97,93 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
+	}
+	userID, ok := v.(string)
+	if !ok {
+		return 0, fmt.Errorf("invalid userinfo response: %v", "invalid ownerClaim value")
+	}
+	if userID != signIn.UserID {
+		return 0, fmt.Errorf("invalid userID: %v", signIn.UserID)
+	}
+
+	v, ok = profile["exp"]
```
jkralik: if `exp` is not set we can consider expiration as forever

jkralik: according to https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4

coap-gateway/service/request.go
```diff
@@ -0,0 +1,16 @@
...
+
+func checkReq(request CoapSignInReq) error {
+	if request.UserID == "" {
+		return status.Errorf(codes.InvalidArgument, "invalid UserId")
+	}
+	if request.AccessToken == "" {
+		return status.Errorf(codes.InvalidArgument, "invalid AccessToken")
+	}
+	return nil
+}
```
jkralik: why it is a separate file -> it is below to sign I think.

coap-gateway/service/refreshToken.go
jkralik: is it use only by signIn ??

coap-gateway/service/service.go
```diff
@@ -211,6 +213,14 @@ func New(ctx context.Context, config Config, logger *zap.Logger) (*Service, erro
 		}
 	}
 
+	provider, err := provider.NewPlgdProvider(config.OAuthDeviceClient,
+		logger,
+		config.Clients.AuthServer.OwnerClaim,
+		"query", "offline", "code")
+	if err != nil {
+		return nil, fmt.Errorf("cannot create device provider: %w", err)
```
jkralik, line 236: add `resourceSubscriber.Close()` before you call return. it is free all allocated resources before.

jkralik:
We need to update the documentation for https://github.com/plgd-dev/cloud/blob/v2/docs/guide/deployment/authorization-server.md and https://github.com/plgd-dev/cloud/blob/v2/docs/guide/deployment/coap-gateway.md

jkralik:
Could we move all action in one PR ? It will be better, because we don't want to provide partial feature.

coap-gateway/service/refreshToken.go
Danielius1922: SignIn uses Unix timestamp in seconds, but yeah I've missed that it's used in one additional place, so I think I'll modify it to use time.Time

coap-gateway/service/request.go
Danielius1922: Will be used by updated `refresh_token` and `sign_off` implementations which currently use a different `checkReq` function, but will be switched when new implementation is done, so moving this to separate file makes sense

coap-gateway/service/refreshToken.go
```diff
@@ -73,7 +72,7 @@ func refreshTokenPostHandler(req *mux.Message, client *Client) {
 	coapResp := CoapRefreshTokenResp{
 		RefreshToken: resp.RefreshToken,
 		AccessToken:  resp.AccessToken,
-		ExpiresIn:    validUntilToExpiresIn(resp.GetValidUntil()),
+		ExpiresIn:    validUntilToExpiresIn(time.Unix(0, resp.GetValidUntil())),
```
jkralik: use `pkgTime.Unix` instead of `time.Unix`.

coap-gateway/service/request.go
```diff
@@ -0,0 +1,14 @@
...
+/// Check that all required request fields are set
+func checkReq(request CoapSignInReq) error {
+	if request.UserID == "" {
+		return fmt.Errorf("invalid UserId")
+	}
+	if request.AccessToken == "" {
+		return fmt.Errorf("invalid AccessToken")
+	}
+	return nil
+}
```
jkralik: rename to checkOAuthRequest

coap-gateway/service/signIn.go
```diff
@@ -94,66 +94,167 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
+	v, ok := jwtToken[expKey]
+	if !ok {
+		return time.Time{}, nil
+	}
+
+	exp, ok := v.(float64) // all integers are float64 in json
+	if !ok {
+		return time.Time{}, fmt.Errorf("invalid userinfo: invalid %v value type", expKey)
+	}
+	return time.Unix(int64(exp), 0), nil
```
jkralik: use `pkgTime.Unix`

coap-gateway/service/signIn.go
```diff
@@ -269,9 +326,50 @@ func signOutPostHandler(req *mux.Message, client *Client, signOut CoapSignInReq)
...
+
+	// fix for iotivity-classic
+	if signOut.UserID == "" {
+		authCurrentCtx, err := client.GetAuthorizationContext()
+		if err != nil {
+			logErrorAndCloseClient(fmt.Errorf("cannot handle sign out: %v", err), coapCodes.InternalServerError)
 			return
 		}
+		signOut.UserID = authCurrentCtx.UserID
+	}
```
jkralik: iotivity-classic is not supported any more, but device be signed so we can use values from AuthContext:
```
authCurrentCtx, _ := client.GetAuthorizationContext()
	userID := signOut.UserID
	deviceID := signOut.DeviceID
	if userID == "" {
		userID = authCurrentCtx.GetUserID()
	}
	if deviceID == "" {
		deviceID = authCurrentCtx.GetDeviceID()
	}
        if accessToken == " {
                accessToken = authCurrentCtx.AccessToken()
        }
```

coap-gateway/service/signIn.go
```diff
@@ -94,66 +94,167 @@ func (client *Client) loadShadowSynchronization(ctx context.Context, deviceID st
...
 }
 
-// https://github.com/openconnectivityfoundation/security/blob/master/swagger2.0/oic.sec.session.swagger.json
-func signInPostHandler(req *mux.Message, client *Client, signIn CoapSignInReq) {
-	resp, err := client.server.asClient.SignIn(req.Context, &pbAS.SignInRequest{
-		DeviceId:    signIn.DeviceID,
-		UserId:      signIn.UserID,
-		AccessToken: signIn.AccessToken,
+/// Get JW Token from oauth server
+func getJWToken(ctx context.Context, service *Service, accessToken string) (map[string]interface{}, error) {
```
jkralik: renamt it to `getUserInfo`

coap-gateway/service/service.go
```diff
@@ -211,6 +213,15 @@ func New(ctx context.Context, config Config, logger *zap.Logger) (*Service, erro
...
 
+	provider, err := provider.NewPlgdProvider(config.OAuthDeviceClient,
+		logger,
+		config.Clients.AuthServer.OwnerClaim,
+		"query", "offline", "code")
+	if err != nil {
+		resourceSubscriber.Close()
+		return nil, fmt.Errorf("cannot create device provider: %w", err)
+	}
+
```
jkralik, line 239: add line with`resourceSubscriber.AddCloseFunc(provider.Close)` - it may solve issue with FD

authorization/pb/service.proto
```diff
@@ -7,12 +7,9 @@ import "github.com/plgd-dev/cloud/authorization/pb/auth.proto";
 option go_package = "github.com/plgd-dev/cloud/authorization/pb;pb";
 
 service AuthorizationService {
-  rpc SignUp (SignUpRequest) returns (SignUpResponse) {}
+	rpc SignUp (SignUpRequest) returns (SignUpResponse) {}
```
ondrejtomcik: From an authorization service point of view, it's no more SignUp or SignOff. It's e.g. RegisterDevice or DeleteDevice. Authorization service now stores only user-device relations and API shall reflect these capabilities.

bundle/run.sh
```diff
@@ -612,6 +612,14 @@ cat /configs/coap-gateway.yaml | yq e "\
   .apis.coap.address = \"${COAP_GATEWAY_UNSECURE_ADDRESS}\" |
   .apis.coap.externalAddress = \"${FQDN}:${COAP_GATEWAY_UNSECURE_PORT}\" |
   .apis.coap.tls.enabled = false |
+  .oauthDeviceClient.provider = \"${DEVICE_PROVIDER}\" |
```
ondrejtomcik: Why you didn't remove these values from authorization-server config - line 474?

coap-gateway/config.yaml
```diff
@@ -18,6 +18,25 @@ apis:
       keyFile: "/secrets/private/cert.key"
       certFile: "/secrets/public/cert.crt"
       clientCertificateRequired: true
+oauthDeviceClient:
```
ondrejtomcik: Please don't forget to remove values from the authorization server config.

ondrejtomcik: Please document newly added values in this repository (https://github.com/plgd-dev/cloud/blob/v2/docs/guide/deployment/coap-gateway.md) and update authorization-server.md as well.

coap-gateway/oauth/oauth.go
```diff
@@ -0,0 +1,88 @@
...
+	switch a {
+	case AuthStyleInHeader:
+		return oauth2.AuthStyleInHeader
+	case AuthStyleInParams:
+		return oauth2.AuthStyleInParams
+	}
+	return oauth2.AuthStyleAutoDetect
+}
+
+//TODO cleanup settings AccessType, ResponseType, ResponseMode, AuthStyle - be carefull it is used by c2c.
```
ondrejtomcik: //TODO when?

coap-gateway/service/config.go
```diff
@@ -16,17 +17,22 @@ import (
 
 //Config represent application configuration
 type Config struct {
-	Log       LogConfig     `yaml:"log" json:"log"`
-	APIs      APIsConfig    `yaml:"apis" json:"apis"`
-	Clients   ClientsConfig `yaml:"clients" json:"clients"`
-	TaskQueue queue.Config  `yaml:"taskQueue" json:"taskQueue"`
+	Log               LogConfig       `yaml:"log" json:"log"`
+	APIs              APIsConfig      `yaml:"apis" json:"apis"`
+	OAuthDeviceClient provider.Config `yaml:"oauthDeviceClient" json:"oauthDeviceClient"`
```
ondrejtomcik: DeviceClient naming doesn't make sense anymore. CoAP gateway handles device connections and logic from the authorization server which was more generic was moved directly to the CoAP gateway. The reasoning for this is that its the CoAP which doesn't suppor OAuth protocol; (or, there is no OAuth Server which would support CoAP). OAuth is defined only for HTTP, so this translation from CoAP OAuth `auth code exchange` to HTTP OAuth `auth code exchange` needs to be done in the CoAP Gateway. 

CoAP is API of the CoAP gateway, and if we want to follow the same convention across all the gateways, this configuration (provider.config) should be part of the apis.coap under authorization. -> apis.coap.authorization, same as you have e.g. for the gRPC.

```yaml
apis:
  coap:
    address: "0.0.0.0:5684"
    externalAddress: ""
    maxMessageSize: 262144
    goroutineSocketHeartbeat: "4s"
    keepAlive:
      timeout: 20s
    blockwiseTransfer:
      enabled: true
      blockSize: "1024"
    tls:
      enabled: true
      caPool: "/secrets/public/rootca.crt"
    authorization:
        provider: "plgd"
        clientID: ""
        clientSecret: ""
        scopes: []
        authorizationURL: ""
        tokenURL: ""
        redirectURL: ""
        http:
             maxIdleConns: 16
...
```

coap-gateway/provider/token.go
```diff
@@ -0,0 +1,13 @@
+package provider
```
ondrejtomcik: It might be misleading, coap gateway provider? what about authorization?

coap-gateway/service/refreshToken.go
```diff
@@ -38,11 +39,10 @@ func validateRefreshToken(req CoapRefreshTokenReq) error {
 	return nil
 }
 
-func validUntilToExpiresIn(v int64) int64 {
-	if v == 0 {
+func validUntilToExpiresIn(validUntil time.Time) int64 {
```
ondrejtomcik, line 65: refresh token request could be handled directly by the new package authorization (now named provider). what do you think?

ondrejtomcik (changes requested):

bundle/run.sh
Danielius1922: I will once all commands are refactored to the new form, remove the stuff from cfg now would break tests for the not-yet-updated code.

Danielius1922:
> Could we move all action in one PR ? It will be better, because we don't want to provide partial feature.

I've changed the target branch to `feature/coap-oauth`, so we can merge this partial implementations to that branch and after the whole feature branch is done I'll merge it to `v2`.

coap-gateway/oauth/oauth.go
Danielius1922: I've copied this class from `authorize` service, so we will have the clean it up there as well. create a story in monday and we'll get to it (eventually)

coap-gateway/service/config.go
Danielius1922: OK, I think it makes sense.

jkralik (commented):
Good job !

jkralik (commented):
Great job :)

docs/guide/deployment/coap-gateway.md
```diff
@@ -137,8 +152,34 @@ CoAP API as specified in the [Open Connectivity Foundation - Device to Cloud Ser
 | `api.coap.tls.certFile` | string | `File path to certificate in PEM format.` | `""` |
 | `api.coap.tls.clientCertificateRequired` | bool | `If true, require client certificate.` | `true` |
 
+#### OAuth2.0 Client for UI and SDK
+
+Configured OAuth2.0 client is used by the mobile application or SDK to request a token used to authorize all calls they execute against other plgd APIs.
```
jkralik: It's used only for devices.

jkralik (approved):
Pls fix the doc

docs/guide/deployment/coap-gateway.md
Danielius1922: `Configured OAuth2.0 client is used by devices to request a token used to authorize all calls they execute against other plgd APIs.`

?

ondrejtomcik: OAuth2.0 Client is used to issuing an authorization code used by the Onboarding Tool during the device provisioning which is then exchanged for an access token during the [SignUp operation](https://plgd.dev/architecture/component-overview/#cloud-registration)

ondrejtomcik: @Danielius1922 could you please update also plantumls in the docs? Adding OAuth2.0 Server against which the code is exchanged and add the CoAP gateway calling authorization-server AddDevice rpc call? Thanks!

ondrejtomcik (approved):
Well done @Danielius1922 ! Good job.

coap-gateway/config.yaml
```diff
@@ -18,6 +18,23 @@ apis:
       keyFile: "/secrets/private/cert.key"
       certFile: "/secrets/public/cert.crt"
       clientCertificateRequired: true
+    authorization:
+      clientID: ""
+      clientSecret: ""
+      scopes: []
+      domain: ""
```
jkralik, line 25: pls rename the `domain` to `authority` - it will be the same as is in  resource-aggregate.

Danielius1922: updated

jkralik (approved):

PR #356, opened 2021-07-29
resource-aggregate/grpc-gateway: implement timeToLive to commands

### LINKED ISSUE #364: Configurable command timeout

jkralik:
commands CREATE,UPDATE,DELETE,RETRIEVE resource and UPDATE metadata
allows to set validUntil as nanoseconds timestamp. After this time
command is considered as expired and cloud don't process it any more.


grpc-gateway/pb/devices.proto
```diff
@@ -76,7 +77,8 @@ message SubscribeToEvents {
     }
     repeated Event event_filter = 1;
     repeated string device_id_filter = 2;
-    repeated string resource_id_filter = 3;
+    repeated string resource_id_filter = 3; // format {deviceID}{href}. eg "ae424c58-e517-4494-6de7-583536c48213/oic/d"
+    bool just_new_events = 4; // if it is set to true, the subscription sends only new events from creation time.
```
ondrejtomcik: This sounds strange in English.
After discussion with @oskarbauer, it is also strange to receive the current state in a first event if you are subscribing to WebSocket events. More clear would be to **not** receive current state in the first event by default and have a boolean `include_current_state`which if set to true (default false), first event containing current_state is sent.

http-gateway/swagger.yaml
```diff
@@ -305,13 +305,14 @@ paths:
         - 'Resources'
       summary: 'Retrieve a resource'
       description: |
-        The plgd cloud builds an up to date [Shadow](https%3A%2F%2Fplgd.dev%2Ffeatures%2Fdevice-shadow%2F) for each published resource. Retrieving it's content using this API returnes an eventually consistent shadow content of the resource. There is a possibility bypass the shadow by setting the query parameter `shadow` to `false.`
+        The plgd cloud builds an up to date [Shadow](https%3A%2F%2Fplgd.dev%2Ffeatures%2Fdevice-shadow%2F) for each published resource. Retrieving it's content using this API returnes an eventually consistent shadow content of the resource. There is a possibility to bypass the shadow by setting the query parameter `shadow` to `false` or by setting set the query parameter interface. In this case, you can set a query parameter validUntil, where you can define the expiration of the command.
```
ondrejtomcik: "or by setting the `interface` query parameter".

"which allows you to set the command expiration."

http-gateway/swagger.yaml
```diff
@@ -516,7 +530,7 @@ paths:
         - 'Resource Events'
       summary: 'Subscribe to events'
       description: |
-        Open a websocket stream and receive events from devices. The contents of filters in the SubscribeToEvents message determine which events will be sent to the stream.
+        Open a websocket stream and receive events from devices. The contents of filters in the SubscribeToEvents message determine which events will be sent to the stream. It will sents current events and then all new filtered events. To bypass current events, set  justNewEvents to true.
```
ondrejtomcik: gramarly

If you're interested in the current state right after subscribing, set the `includeCurrentState` query parameter to `true`.

http-gateway/swagger.yaml
```diff
@@ -1359,7 +1376,14 @@ components:
...
       name: timestamp
       in: query
-      description: 'Filter by Unix time in nanoseconds'
+      description: 'Filter by Unix nanoseconds timestamp'
+      schema:
+        type: integer
+        format: int64
+    validUntil:
+      name: validUntil
+      in: query
```
ondrejtomcik, line 1390: what is the default?

http-gateway/swagger.yaml
```diff
@@ -1359,7 +1376,14 @@ components:
...
+      description: 'Defines expiration of command in Unix nanoseconds timestamp. 0 means never.'
```
ondrejtomcik: Command expiration defined as Unix timestamp in nanoseconds.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -93,8 +93,22 @@ func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataSnapshotTaken(ctx cont
 	return nil
 }
 
-func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataUpdatePending(ctx context.Context, updatePending *DeviceMetadataUpdatePending) error {
+func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataUpdatePending(ctx context.Context, updatePending *DeviceMetadataUpdatePending, isCmd bool) error {
+	now := time.Now()
+	if !updatePending.IsValid(now) {
```
ondrejtomcik: Aggregate shall not validate commands. Command validation is done in a previous layer. It doesn't belong here.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -93,8 +93,22 @@ func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataSnapshotTaken(ctx cont
...
+		if isCmd {
```
ondrejtomcik: It's always a command. If you do the validation at a proper place, you can remove this unreadable hack :)

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -93,8 +93,22 @@ func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataSnapshotTaken(ctx cont
...
+			return status.Errorf(codes.InvalidArgument, "valid_until(`%v`) is less than now(`%v`)", updatePending.ValidUntilTime(), now)
+		}
+		e.DeviceId = updatePending.GetDeviceId()
+		e.EventMetadata = updatePending.GetEventMetadata()
+		// for events from eventstore we do nothing
+		return nil
+	}
 	for _, event := range e.GetUpdatePendings() {
+		if !event.IsValid(now) {
+			continue
```
ondrejtomcik, line 106: Don't forget to document that expired pending commands are removed only when the new snapshot is created.

resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -80,8 +80,42 @@ func (e *ResourceStateSnapshotTaken) CheckInitialized() bool {
...
+type resourceValidUntilValidator interface {
+	ValidUntilTime() time.Time
+	IsValid(now time.Time) bool
+	GetEventMetadata() *EventMetadata
+	GetAuditContext() *commands.AuditContext
+	GetResourceId() *commands.ResourceId
+}
+
+func (e *ResourceStateSnapshotTaken) processValidUntil(v resourceValidUntilValidator, isCmd bool, now time.Time) (bool, error) {
+	if !v.IsValid(now) {
```
ondrejtomcik: Validation doesn't belong to aggregate.

resource-aggregate/events/resourceUpdatePending.go
```diff
@@ -56,3 +56,15 @@ func (e *ResourceUpdatePending) CheckInitialized() bool {
 		e.GetAuditContext() != nil &&
 		e.GetEventMetadata() != nil
 }
+
+func (e *ResourceUpdatePending) ValidUntilTime() time.Time {
+	return pkgTime.Unix(0, e.GetValidUntil())
+}
+
+func (e *ResourceUpdatePending) IsValid(now time.Time) bool {
+	v := e.ValidUntilTime()
```
ondrejtomcik: Same code in 4 files .

resource-aggregate/pb/commands.proto
```diff
@@ -461,6 +465,7 @@ message UpdateDeviceMetadataRequest {
         ConnectionStatus status = 3;
         ShadowSynchronization shadow_synchronization = 4;
     };
+    int64 valid_until = 99;  // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano) when command is considered as expired. 0 means forever.
```
ondrejtomcik: Everywhere else you incremented by one, here you jumped to 99. Please fix.

resource-aggregate/pb/events.proto
```diff
@@ -16,69 +16,72 @@ message EventMetadata {
...
-    EventMetadata event_metadata = 5;
+    AuditContext audit_context = 100;
+    EventMetadata event_metadata = 101;
 }
 
 message ResourceDeletePending {
     ResourceId resource_id = 1;
     AuditContext audit_context = 100;
     EventMetadata event_metadata = 101;
+    int64 valid_until = 102;  // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano) when pending event is considered as expired. 0 means forever.
```
ondrejtomcik: field numbers in the range 1 through 15 take one byte to encode, while 15+ takes 2 bytes. 1-15 shall be used  for **very frequently occurring message elements**. validUntil, auditContext and eventMetadata are always present and shall be imho always in range 1-15.

resource-directory/service/deviceMetadataProjection.go
```diff
@@ -43,22 +44,25 @@ func (p *deviceMetadataProjection) InitialNotifyOfDeviceMetadata(ctx context.Con
...
+}
+
+func (p *deviceMetadataProjection) onDeviceMetadataUpdatePendingLocked(ctx context.Context, do func(ctx context.Context, createPending *events.DeviceMetadataUpdatePending) error) error {
 	log.Debugf("deviceMetadataProjection.onDeviceMetadataUpdatePendingLocked %v", p.data.GetUpdatePendings())
 	var errors []error
-	for _, u := range p.data.GetUpdatePendings() {
-		err := p.subscriptions.OnDeviceMetadataUpdatePending(ctx, u)
+	now := time.Now()
+	for _, ev := range p.data.GetUpdatePendings() {
+		if !ev.IsValid(now) {
```
ondrejtomcik: if ev.IsExpired

resource-aggregate/events/deviceMetadataUpdatePending.go
```diff
@@ -55,3 +55,15 @@ func (e *DeviceMetadataUpdatePending) CheckInitialized() bool {
 		e.GetAuditContext() != nil &&
 		e.GetEventMetadata() != nil
 }
+
+func (e *DeviceMetadataUpdatePending) ValidUntilTime() time.Time {
+	return pkgTime.Unix(0, e.GetValidUntil())
+}
+
+func (e *DeviceMetadataUpdatePending) IsValid(now time.Time) bool {
```
ondrejtomcik: IsExpired

resource-aggregate/events/resourceCreatePending.go
```diff
@@ -54,3 +54,15 @@ func (e *ResourceCreatePending) CheckInitialized() bool {
 		e.GetAuditContext() != nil &&
 		e.GetEventMetadata() != nil
 }
+
+func (e *ResourceCreatePending) ValidUntilTime() time.Time {
+	return pkgTime.Unix(0, e.GetValidUntil())
+}
+
+func (e *ResourceCreatePending) IsValid(now time.Time) bool {
```
ondrejtomcik: IsExpired

resource-aggregate/events/resourceDeletePending.go
```diff
@@ -52,3 +52,15 @@ func (e *ResourceDeletePending) CheckInitialized() bool {
 		e.GetAuditContext() != nil &&
 		e.GetEventMetadata() != nil
 }
+
+func (e *ResourceDeletePending) ValidUntilTime() time.Time {
+	return pkgTime.Unix(0, e.GetValidUntil())
+}
+
+func (e *ResourceDeletePending) IsValid(now time.Time) bool {
```
ondrejtomcik: IsExpired

resource-aggregate/events/resourceRetrievePending.go
```diff
@@ -54,3 +54,15 @@ func (e *ResourceRetrievePending) CheckInitialized() bool {
 		e.GetAuditContext() != nil &&
 		e.GetEventMetadata() != nil
 }
+
+func (e *ResourceRetrievePending) ValidUntilTime() time.Time {
+	return pkgTime.Unix(0, e.GetValidUntil())
+}
+
+func (e *ResourceRetrievePending) IsValid(now time.Time) bool {
```
ondrejtomcik: IsExpired

ondrejtomcik (changes requested):

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: It's bellows here because, event can be expired during processing eventstore.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: No. It's called during load events from eventstore.

resource-aggregate/events/resourceStateSnapshotTaken.go
jkralik: It's bellows here because events can be expired during processing eventstore.

resource-aggregate/pb/commands.proto
jkralik: From my point of view, we need to hold the same number for the same type among all events.

After that user can define his own message for parse only sub-values.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
ondrejtomcik: I mean event. At this you are aggregating events. Command is supposed to be already validated and event created. It doesn’t belong here.

jkralik: this is part where we validate events -> eg duplicit correlation id, snapshot always must be valid

ondrejtomcik: Where do we validate duplicit correlation id of the command? Correlation id is created by us, we do not validate if such a correlation id exist already. Only matching is done when response comes, but this  requires loading of events and finding if one matching that correlation id is there. ValidUntil is static field independent of current state. 
Your argument that it can expire during events aggregating is not well placed because same can happen with the token. Additionally, we should always require at least +100ms because event can be pushed to nats but at the moment its processed by coap gateway it might be already expired. 

As I am thinking about it, the best would be to not use unix timestamp - validUntil but ttl, defined as number of ms command is valid from counted from the moment command is accepted. Command accept response might contain value, unix timestamp when it expires. This would be much better design and you don’t need any validation if we require that it must be 0 or higher than 100ms.

jkralik: Validation is processed by snapshot by this handlers https://github.com/plgd-dev/cloud/pull/356/files#diff-c0979327025ce3f41a9ec7eb64cafdcedacb80760b1981bdbbe764b31f593616R112

I don't agree with TTL. The user makes command and he is responsible when the command expires. With `ttl` we must send a response with `validUntil`. And maybe at that time, the command is invalid.

ondrejtomcik: Command request confirmation (http / grpc response) contains valid until if ttl was valid. What usecase do you have in mind that at the time of this response the command is already expired? 

Http create command with ttl -> command validation (ttl == 0 || ttl > 100) -> create pending command event -> return eventmetadata time + ttl in confirmation response to client. 

This design is much less error prone.

jkralik: No true -> eg browser created command with ttl 10s -> and send request, the response comes after 10s, but browser considers it as canceled(because it takes 10s), with validUntil backend recognize such state pending event is not processed.

jkralik: validUntil is deterministic, ttl is floating

ondrejtomcik: ValidUntil requires time synchronization. You need to know the server time and seconds drift can for sure occur and of course, if there is drift 3 seconds, and user set validUntil to his time+ 2s command might be immediately rejected. That’s why absolute time is a bad idea. 

If clients sets ttl to 10s its up to backend to cancel it in 10s + processing. It shouldnt be anymore responsibility of the client if this query param is there.

jkralik: cloud must have synchronized time via NTP - for aws/azure they are responsible for it.

| If clients sets ttl to 10s its up to backend to cancel it in 10s + processing. It shouldnt be anymore responsibility of the client if this query param is there.

But the client doesn't want to execute the command when he cancel operation! For your scenario it same as you set validUntil=0. You don't care when the command is executed.

ondrejtomcik: Ntp synchronization is not enough. Ttl is from clients usually send as timeout in seconds exactly to prevent such an errors and to not rely on some time synchronization. Using absolute times calculated on the client side leads to errors. 

If the client set ttl and cancels earlier, it should not cancel the command. It should only cancel waiting for an response but the command is already registered.

jkralik: I can only add ttl but validUntil will stay as is.

resource-aggregate/pb/events.proto
jkralik: 2bytes is not big deal for sharing id.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
ondrejtomcik: Why we should have both of them? They serve the same purpose.

jkralik: I still see 2 use cases:
- user wants to define absolute time (validUntil), it is from created request by user
- user wants to define drift time(ttl), validUntil will be defined by resource-aggregate

jkralik: Eg GetResourceFromDevice is a command where I want to use valid until - if I cancel by timeout command I don't want to get any more data from device.

ondrejtomcik: In your example, you can use just ttl and send cancel command - new api. Async=False query parameter we are going to introduce is about **synchronously waiting for a response**. Canceling waiting != canceling command. Command cancelation shall not be done by cancelling the session. 

regarding your user usecases, user have just one usecase. Set the time when the command becomes invalid. You described two technical solutions to do the same, not two user scenarios. 🙂

jkralik: ok so I'm for first scenario 😄

ondrejtomcik: You mean first technical solution. Me not. So let’s clarify with API consumers. I will write an email to few of them this evening. Until we have a feedback, let’s put this MR aside. 

@Danielius1922 please provide also your input. Thanks.

ondrejtomcik: > eg browser created command with ttl 10s -> and send request, the response comes after 10s, but browser considers it as canceled(because it takes 10s), with validUntil backend recognize such state pending event is not processed.

Same can happen with the validUntil. Device comes online second before expiration and command is executed. Execution takes 2 seconds. ValidUntil is by client considered as expired, but it's being processed by the device. :)

> But the client doesn't want to execute the command when he cancel operation! 

If client sent the command, this command is registered and event commandPending created. If client cancels HTTP session on which he is synchronously waiting for a response, this is not canceling the pending command. Pending command stays there.

jkralik: Fixed

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: https://github.com/plgd-dev/cloud/blob/validUntilCommands/resource-aggregate/pb/events.proto#L114

grpc-gateway/pb/updateDeviceMetadata.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (req *UpdateDeviceMetadataRequest) ToRACommand(ctx context.Context) (*commands.UpdateDeviceMetadataRequest, error) {
+	correlationUUID, err := uuid.NewV4()
+	if err != nil {
+		return nil, err
+	}
+	connectionID := ""
+	peer, ok := peer.FromContext(ctx)
+	if ok {
+		connectionID = peer.Addr.String()
```
Danielius1922, line 19: in case `ok == false` we create a `CommandMetadata` object with empty `ConnectionId`, is that OK?

grpc-gateway/pb/updateDeviceMetadata.pb.go
```diff
@@ -74,6 +74,7 @@ type UpdateDeviceMetadataRequest struct {
 
 	DeviceId              string                                            `protobuf:"bytes,1,opt,name=device_id,json=deviceId,proto3" json:"device_id,omitempty"`
 	ShadowSynchronization UpdateDeviceMetadataRequest_ShadowSynchronization `protobuf:"varint,2,opt,name=shadow_synchronization,json=shadowSynchronization,proto3,enum=ocf.cloud.grpcgateway.pb.UpdateDeviceMetadataRequest_ShadowSynchronization" json:"shadow_synchronization,omitempty"`
+	TimeToLive            int64                                             `protobuf:"varint,5,opt,name=timeToLive,proto3" json:"timeToLive,omitempty"` // command validity in nanoseconds. 0 means forever and minimal valus is 100000000 (100ms).
```
Danielius1922: typo: valus -> value

grpc-gateway/pb/updateDeviceMetadata.proto
```diff
@@ -13,6 +13,7 @@ message UpdateDeviceMetadataRequest{
     }
     string device_id = 1;
     ShadowSynchronization shadow_synchronization = 2;
+    int64 timeToLive = 5;  // command validity in nanoseconds. 0 means forever and minimal valus is 100000000 (100ms).
```
Danielius1922: why `= 5` and not `3`?

Danielius1922 (approved):

grpc-gateway/pb/updateDeviceMetadata.go
jkralik: Yes. it's ok. We just fill it in when we have such information.

http-gateway/swagger.yaml
```diff
@@ -1359,10 +1380,18 @@ components:
...
       in: query
-      description: 'Filter by Unix time in nanoseconds'
+      description: 'Filter by Unix nanoseconds timestamp'
+      schema:
+        type: integer
+        format: int64
+    timeToLive:
+      name: timeToLive
+      in: query
+      description: 'Command expiration in nanoseconds. 0 means forever and 100ms is minimal value.'
```
ondrejtomcik, line 1391: you can specify in swagger also min afaik.

http-gateway/swagger.yaml
```diff
@@ -1359,10 +1380,18 @@ components:
...
       schema:
         type: integer
         format: int64
```
ondrejtomcik, line 1394: didn't we discuss that it will be string where you can specify also unit?

resource-aggregate/pb/events.proto
ondrejtomcik: But why to not have one. There is extremely low chance that we will occupy additional 15-1+3 1byte positions.

ondrejtomcik:
There are some lint checks failing.

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -94,7 +94,17 @@ func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataSnapshotTaken(ctx cont
 }
 
 func (e *DeviceMetadataSnapshotTaken) HandleDeviceMetadataUpdatePending(ctx context.Context, updatePending *DeviceMetadataUpdatePending) error {
+	now := time.Now()
+	if updatePending.IsExpired(now) {
+		e.DeviceId = updatePending.GetDeviceId()
+		e.EventMetadata = updatePending.GetEventMetadata()
```
ondrejtomcik, line 100: why you are not storing e.AuditContext?

resource-aggregate/service/grpcApi.go
```diff
@@ -201,9 +201,19 @@ func (r RequestHandler) UpdateResource(ctx context.Context, request *commands.Up
 	if err != nil {
 		log.Errorf("cannot publish resource content update events: %v", err)
 	}
+
+	var validUntil int64
+	for _, e := range events {
```
ondrejtomcik, line 203: Why iterating and not taking events[len(events)-1] ?

resource-aggregate/service/grpcApi.go
```diff
@@ -201,9 +201,19 @@ func (r RequestHandler) UpdateResource(ctx context.Context, request *commands.Up
 	if err != nil {
 		log.Errorf("cannot publish resource content update events: %v", err)
 	}
+
```
ondrejtomcik, line 201: Why you don't validate on line 189 if ttl is >= 100ms?

ondrejtomcik: same for other calls

resource-aggregate/service/grpcApi.go
ondrejtomcik: same for other calls

ondrejtomcik:
@jkralik FYI https://github.com/plgd-dev/cloud/issues/364#issuecomment-892997316

http-gateway/swagger.yaml
jkralik: I change it because for developers it is easier to use numbers than concatenate strings.

http-gateway/swagger.yaml
jkralik: Fixed

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: For the same reason as `resourceLinksSnapshotTaken` - snapshot doesn't contain audit context because it is produced by our system, not by the user.

resource-aggregate/service/grpcApi.go
jkralik: in cancellation MR it can create a snapshot after the event so I cannot use it. And in the worst case, it contains 2 events.

resource-aggregate/service/grpcApi.go
jkralik: Fixed

resource-aggregate/pb/events.proto
jkralik: Fixed

jkralik:
> There are some lint checks failing.

Fix of lint doesn't below here - it is now in each PR. So we need to fix it in separate PR

ondrejtomcik (approved):

PR #361, opened 2021-07-31
pkg: add error checks

Danielius1922:


pkg/sync/task/queue/queue.go
```diff
@@ -56,19 +56,20 @@ func (q *Queue) popQueue() func() {
...
 			task := q.popQueue()
 			if task == nil {
 				return
 			}
 			task()
 		}
-	})
+	}); err != nil {
+		return err
+	}
```
jkralik: we just try to submit action - so when it fails, it means that we reach the limit of goroutines, and the task will be processed by another goroutine.

pls change it to: `_ = q.goPool.Submit(...`

jkralik (approved):

PR #381, opened 2021-08-04
resource-aggregate: add Delete call to eventstore

### LINKED ISSUE #365: Force device deletion directly from the plgd Cloud
Being able to delete the device from the plgd cloud if it's online, or cleanup the list of devices which are already disconnected and will never connect again.

Devices can be currently deleted from the plgd only using the onboarding tool, on the local network. But even when using this path it's not guaranteed that the device will be removed from the plgd cloud. If the device is disconnected from the plgd cloud (network issue) but OBT can reach it and reset, this device will never connect back to plgd cloud but it's shadow will be still in the list. User needs a possibility to delete stale devices or trigger deregistration of online device directly from the plgd cloud.

None are available

Danielius1922:
Add Delete call to eventstore which deletes all documents with given groupIds.


resource-aggregate/cqrs/eventstore/mongodb/delete.go
```diff
@@ -0,0 +1,59 @@
...
+	col := s.client.Database(s.DBName()).Collection(getEventCollectionName())
+
+	var deletedDeviceIds []string
+	var errors []error
+	for deviceId := range deviceIds {
+		res, err := col.DeleteMany(ctx, bson.M{
+			groupIDKey: deviceId,
+		})
+		if err != nil {
+			errors = append(errors, fmt.Errorf("failed to delete documents with groupid(%v): %v", deviceId, err))
```
jkralik: pls return the error - it happens when something bad happens.

Danielius1922: fixed

jkralik (approved):

grpc-gateway/config.yaml
```diff
@@ -35,6 +35,19 @@ apis:
           certFile: "/secrets/public/cert.crt"
           useSystemCAPool: false
 clients:
+  authorizationServer:
+    ownerClaim: sub
```
jkralik: do you need ownerClaim ?

authorization/service/deleteDevices.go
```diff
@@ -0,0 +1,70 @@
...
+
+	deviceIds := getUniqueDeviceIds(request.DeviceIds)
+	if len(deviceIds) == 0 {
+		return nil, log.LogAndReturnError(status.Errorf(codes.InvalidArgument, "cannot delete devices: invalid DeviceIds"))
+	}
+
+	// TODO validate jwt token -> only jwt token is supported
+
+	var deletedDeviceIds []string
+	for _, deviceId := range request.DeviceIds {
```
jkralik: if deviceIDs are empty you don't proceed any deletein auth service. but in ra you delete all resources of user.

resource-aggregate/service/deleteDevices.go
```diff
@@ -0,0 +1,55 @@
...
+	return deviceIds.ToSlice()
+}
+
+func (r RequestHandler) DeleteDevices(ctx context.Context, request *commands.DeleteDevicesRequest) (*commands.DeleteDevicesResponse, error) {
+	deviceIds := getUniqueDeviceIdsFromDeleteRequest(request)
+	if len(deviceIds) == 0 {
+		return nil, log.LogAndReturnError(kitNetGrpc.ForwardErrorf(codes.InvalidArgument, "cannot delete devices: invalid request"))
+	}
+
+	owner, ownedDevices, err := r.getOwnedDevices(ctx, deviceIds)
```
jkralik, line 35: if deviceIds is not set, you delete all devices at RA but in auth, you don't delete any if the request doesn't contain deviceIds

jkralik (changes requested):
pls fix inconsistency of delete between RA and AUTH service for empty deviceIds.

grpc-gateway/service/deleteDevices.go
```diff
@@ -0,0 +1,81 @@
...
+				log.Errorf("failed to delete device('%v') in ResourceAggregate", deviceId)
+			}
+		}
+	}
+
+	if len(deleted) == 0 {
+		return &pb.DeleteDevicesResponse{
+			DeviceIds: deleted,
+		}, nil
+	}
```
jkralik: if `deleteAllOwned==true` and `len(deleted) == 0` then we want to call `r.authorizationClient.DeleteDevices(ctx, &cmdAS)`

jkralik: we always want to delete all devices from authorizationClient with same argument as in `cmdRA.DeviceIds`

grpc-gateway/config.yaml
Danielius1922: ownerClaim removed

coap-gateway/service/signOff.go
```diff
@@ -118,14 +118,18 @@ func signOffHandler(req *mux.Message, client *Client) {
 	}
 
 	ctx = kitNetGrpc.CtxWithToken(ctx, signOffData.accessToken)
-
-	if _, err := client.server.asClient.DeleteDevice(ctx, &pb.DeleteDeviceRequest{
-		DeviceId: signOffData.deviceID,
-		UserId:   signOffData.userID,
-	}); err != nil {
+	resp, err := client.server.asClient.DeleteDevices(ctx, &pb.DeleteDevicesRequest{
```
jkralik: maybe we need to call deleteDevices to RA

Danielius1922: Yes, I think it makes sense. Fixed.

jkralik (approved):

authorization/client/userDevices.go
```diff
@@ -169,6 +169,19 @@ func (d *UserDevicesManager) UpdateUserDevices(ctx context.Context, userID strin
 	return userDevices[userID], nil
 }
 
+func (d *UserDevicesManager) ChecksUserDevices(userID string, deviceIDs []string) bool {
```
ondrejtomcik, line 172: Just came to my mind, should we really name it User? Isn't it the owner? Also RPC calls, AddUserDevices. It's assigning to the owner, and the user doesn't have to necessarily be an end-user, or? @Danielius1922 , @jkralik what do you think?

Danielius1922: `Owner` seems better, I think @jkralik suggested something similar: renaming GetUserDevices to GetOwnerDevices. If it better expresses what it does I think we should rename it, though we should rename it in all places and not just in one or two.

jkralik (approved):

authorization/client/userDevices.go
ondrejtomcik: Discussed, we'll do it but in separate MR.

authorization/service/addDevice.go
```diff
@@ -54,11 +54,12 @@ func (s *Service) AddDevice(ctx context.Context, request *pb.AddDeviceRequest) (
 	defer tx.Close()
 
 	owner := request.UserId
-	token, err := grpc_auth.AuthFromMD(ctx, "bearer")
-	if err != nil {
-		uid, err := grpc.ParseOwnerFromJwtToken(s.ownerClaim, token)
-		if err == nil && uid != serviceOwner {
-			owner = uid
+	if owner == "" {
```
ondrejtomcik, line 61: This will be refactored later? As we discussed owner will be taken only from the token.

authorization/service/deleteDevices.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (s *Service) publishDevicesUnregistered(ctx context.Context, owner string, deviceIDs []string) error {
+	v := events.Event{
+		Type: &events.Event_DevicesUnregistered{
+			DevicesUnregistered: &events.DevicesUnregistered{
+				Owner:     owner,
+				DeviceIds: deviceIDs,
+				AuditContext: &events.AuditContext{
+					UserId: owner,
```
ondrejtomcik, line 52: This should be always the "sub" claim. Config.ownerClaim can be changed from "sub" to something else and in such a case, we lose the information on who did this change.

authorization/service/deleteDevices.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil
+}
+
+// DeleteDevices removes a devices from user.
+func (s *Service) DeleteDevices(ctx context.Context, request *pb.DeleteDevicesRequest) (*pb.DeleteDevicesResponse, error) {
+	tx := s.persistence.NewTransaction(ctx)
+	defer tx.Close()
+
+	owner := request.UserId
+	if owner == "" {
```
ondrejtomcik, line 82: Same applies here.

authorization/service/deleteDevices.go
```diff
@@ -0,0 +1,135 @@
...
+				owner = uid
+			}
+		}
+	}
+
+	if owner == "" {
+		return nil, log.LogAndReturnError(status.Errorf(codes.InvalidArgument, "cannot delete devices: invalid UserId"))
+	}
+
+	// TODO validate jwt token -> only jwt token is supported
```
ondrejtomcik: Is this in monday? Which task?

grpc-gateway/config.yaml
```diff
@@ -35,6 +35,18 @@ apis:
           certFile: "/secrets/public/cert.crt"
           useSystemCAPool: false
 clients:
+  authorizationServer:
```
ondrejtomcik, line 38: Please update also deployment config which is still part of this repo.

ondrejtomcik: @rafajpet FYI

http-gateway/swagger.yaml
```diff
@@ -49,6 +49,42 @@ paths:
           $ref: '#/components/responses/badRequest'
         401:
           $ref: '#/components/responses/unauthorized'
+    delete:
+      tags:
+        - 'Devices'
+      summary: 'Delete owned devices'
```
ondrejtomcik: you can skip word owned. It might raise the question "what is owned device", "if there can be connected not owned device", and so on. Each device connected to the plgd cloud is owned.

resource-aggregate/.plantuml
```diff
@@ -296,4 +296,28 @@ ResourceAggregate --> Bus : DeviceMetadataUpdated (Canceled=true)
...
+entity "OCF Server" as Server
+participant "Gateway" as Gateway
+participant "Resource Aggregate" as ResourceAggregate
+participant "Authorization Server" as AuthServer
+
+activate ResourceAggregate
+activate Server
+activate Gateway
+activate AuthServer
+Server -> Gateway : DeleteDevicesRequest
```
ondrejtomcik: Is it really a device sending the delete request? The device sends SignOff. This example should be about the client, nope?

resource-aggregate/.plantuml
```diff
@@ -296,4 +296,28 @@ ResourceAggregate --> Bus : DeviceMetadataUpdated (Canceled=true)
...
+@startuml
+title "Delete Devices"
+
+entity "OCF Server" as Server
+participant "Gateway" as Gateway
+participant "Resource Aggregate" as ResourceAggregate
+participant "Authorization Server" as AuthServer
+
+activate ResourceAggregate
+activate Server
```
ondrejtomcik: activation represents processing of the request. You have to do it step by step. :)

resource-aggregate/.plantuml
```diff
@@ -296,4 +296,28 @@ ResourceAggregate --> Bus : DeviceMetadataUpdated (Canceled=true)
...
+activate Gateway
+activate AuthServer
+Server -> Gateway : DeleteDevicesRequest
+Gateway-> ResourceAggregate : DeleteDevicesRequest
```
ondrejtomcik: Could you please include also EventBus into this diagram and visualize events which these components, RA, AS are sending?

resource-aggregate/.plantuml
```diff
@@ -296,4 +296,28 @@ ResourceAggregate --> Bus : DeviceMetadataUpdated (Canceled=true)
...
+ResourceAggregate -> Gateway : DeleteDevicesResponse
+Gateway -> AuthServer : DeleteDevicesRequest
+AuthServer -> Gateway : DeleteDevicesResponse
+Gateway -> Server : DeleteDevicesResponse
```
ondrejtomcik: Could you please include this diagram also in the plgd.dev docs, `archtiecture/component-overview/CoAP Gateway` with header Device removed by the user from the plgd Cloud ?

And make it more detailed, including interaction with the device after delete:
example:
```
...
CGW -> Server: Disconnect
destroy Server
deactivate CGW
Server -> CGW: SignUp
activate Server
activate CGW
CGW -> AS: ??? question on chat
...
CGW -> Server: Unauthorized
CGW -> Server: Disconnect
destroy Server
deactivate CGW
Server -> Server: Remove cloud credentials
```

ondrejtomcik: This MR can be closed ofc, this comment is not blocking this one :)

authorization/service/addDevice.go
Danielius1922: Yes, it must be done later, because if I remove this it will break tests. We need to develop a mechanism with inserts a custom userId into a token and update the tests, not sure if there is a story for this in monday

authorization/service/deleteDevices.go
Danielius1922: Shouldn't it be removed from configuration then if it shouldn't be changed?

authorization/service/deleteDevices.go
Danielius1922: Hmm copy & paste from original implementation in authorization/service/deleteDevice.go, not sure if it is relevant anymore...?

grpc-gateway/config.yaml
Danielius1922: Could you please link it? I don't know what you mean.

authorization/service/deleteDevices.go
ondrejtomcik: No, because the `event.Owner` should match `ownerClaim` claim value. `AuditContext` always "sub". Reason:
`ownerClaim` is set to tenantId - organization identifier. That means that the device is owned by the organization and any employee of the organization can retrieve it. If you pass the same id (tenantId) to the AuditContext, all you can track back is that this organization deleted it. But you don't know which bad guy did this mistake, because you didn't take his userId, which is always in "sub" claim :) So it's a good feature for bad guys, but not for bad security managers :D

authorization/service/deleteDevices.go
ondrejtomcik: Validation of JWT is done somewhere else, right? Then you can delete it :)

resource-aggregate/.plantuml
Danielius1922: Not sure what you mean here. This actually deletes all events for given devices from the RA database and all devices from the Auth database ...?

grpc-gateway/config.yaml
ondrejtomcik: https://github.com/plgd-dev/cloud/blob/v2/docs/guide/deployment/grpc-gateway.md

authorization/service/deleteDevices.go
```diff
@@ -0,0 +1,135 @@
...
+
+		err = tx.Delete(deviceId, owner)
+		if err != nil {
+			return nil, log.LogAndReturnError(status.Errorf(codes.NotFound, "cannot delete device('%v'): not found", deviceId))
+		}
+
+		deletedDeviceIds = append(deletedDeviceIds, deviceId)
+	}
+
+	if err := s.publishDevicesUnregistered(ctx, owner, deletedDeviceIds); err != nil {
```
ondrejtomcik, line 130: I am not really sure if we can send this event. At the moment we introduce sharing of devices, or customers will implement it in their own auth service, it's impossible to distinguish if the device was really deleted or if sharing for user B was stopped, but it's still in the system for user A. Plan was to send this event if device sharing is stopped for user B by user A for example. But then, coap gateway listening on these events cannot decide if it should disconnect it or not. Does this explanation make sense?
@Danielius1922 @jkralik

authorization/service/deleteDevices.go
Danielius1922: In the case that ownerClaim != 'sub' where will both claims be. In the JTW? Is it guaranteed to always have both non-empty? Or what should I do when 'sub' or custom claim is empty?

ondrejtomcik: All gateways have to verify 
if ownerClaim == "sub"
  jwt.sub must be present
if ownerClaim == "abc"
  jwt.sub && jwt.abc must be present

ondrejtomcik: Will be covered in the separate task. @Danielius1922 please add todo at this line

authorization/service/deleteDevices.go
ondrejtomcik: todo comment

grpc-gateway/config.yaml
Danielius1922: grpc-gw documentation updated

authorization/service/deleteDevices.go
Danielius1922: TODO added

authorization/service/deleteDevices.go
Danielius1922: TODO added

ondrejtomcik (approved):

PR #386, opened 2021-08-06
coap-gateway: fix auth context in refresh token

Danielius1922:
Don't cache authorization context if device is not signed in.

### DIFF
--- a/coap-gateway/service/refreshToken.go
+++ b/coap-gateway/service/refreshToken.go
@@ -123,18 +123,20 @@ func refreshTokenPostHandler(req *mux.Message, client *Client) {
 		return
 	}
 
-	authCtx := authorizationContext{
-		DeviceID:    refreshToken.DeviceID,
-		UserID:      owner,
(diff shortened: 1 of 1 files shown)


jkralik (approved):

coap-gateway/service/refreshToken.go
```diff
@@ -123,18 +123,20 @@ func refreshTokenPostHandler(req *mux.Message, client *Client) {
...
 	}
 
-	authCtx := authorizationContext{
-		DeviceID:    refreshToken.DeviceID,
-		UserID:      owner,
-		AccessToken: token.AccessToken,
-		Expire:      validUntil,
-	}
-	client.SetAuthorizationContext(&authCtx)
+	if _, err := client.GetAuthorizationContext(); err != nil {
```
jkralik, line 126: err == nil :D

Danielius1922: bah, indeed

PR #388, opened 2021-08-06
coap-gateway: device access token must be JWT

jkralik:
dropping reference token


coap-gateway/service/signOff_test.go
```diff
@@ -15,16 +15,16 @@ func TestSignOffHandler(t *testing.T) {
...
 	err := co.Close()
 	require.NoError(t, err)
 
 	tbl := []testEl{
 		{"Bad query", input{coapCodes.DELETE, `{}`, []string{"di=%"}}, output{coapCodes.BadOption, "invalid URL escape", nil}, true},
 		{"Bad request (no userId)", input{coapCodes.DELETE, `{}`, []string{"di=" + CertIdentity}}, output{coapCodes.BadRequest, "invalid user id", nil}, true},
-		{"Bad request (invalid userId)", input{coapCodes.DELETE, `{}`, []string{"di=" + CertIdentity, "accesstoken=" + signUpResp.AccessToken, "uid=0"}}, output{coapCodes.InternalServerError, "invalid ownerClaim", nil}, true},
 		{"Bad request (missing access token)", input{coapCodes.DELETE, `{}`, []string{"di=" + CertIdentity, "uid=0"}}, output{coapCodes.BadRequest, `invalid access token`, nil}, true},
-		{"Deleted0", input{coapCodes.DELETE, `{}`, []string{"di=" + CertIdentity, "accesstoken=" + signUpResp.AccessToken, "uid=" + signUpResp.UserID}}, output{coapCodes.Deleted, nil, nil}, false},
+		// not supported because JWT has length more than 255 bytes
```
Danielius1922, line 26: This shouldn't be problem in iotivity plgd branch, or is there some other layer that fails with token longer than 255b?

pkg/security/jwt/validator.go
```diff
@@ -34,6 +35,26 @@ func (v *Validator) Parse(token string) (jwt.MapClaims, error) {
...
+
+	jwtKeyfunc := func(token *jwt.Token) (interface{}, error) {
+		return v.keys.GetOrFetchKeyWithContext(ctx, token)
+	}
+
+	t, err := jwt.Parse(token, jwtKeyfunc)
+	if t == nil {
+		return nil, fmt.Errorf("could not parse token: %w", err)
+	}
+	c := t.Claims.(jwt.MapClaims)
```
Danielius1922: This seems a bit weird, shouldn't it be something like?

```
c, ok := t.Claims.(jwt.MapClaims)
if !ok {
  return error
}

**Review by Danielius1922 — APPROVED** [review]

File: coap-gateway/service/signOff_test.go (hunk 1 above)
Comment (jkralik) [reply coap-gateway/service/signOff_test.go:26]: this test doesn't use iotivity-lite device - it uses go-coap/tcp/client which emulates device

File: coap-gateway/authorization/plgd.go [hunk 3]
```diff
@@ -76,7 +76,7 @@ func (p *PlgdProvider) Exchange(ctx context.Context, authorizationProvider, auth
 	}
 
 	oauthClient := p.OAuth2.Client(ctx, token)
-	resp, err := oauthClient.Get(p.UserInfoURL)
+	resp, err := oauthClient.Get(p.OpenID.UserInfoURL)
```
ondrejtomcik: You don't need UserInfo at all. If we use JWT tokens, everything this endpoint returns is part of the JWT. ownerClaim is what contains ownerID (in this code userID [rename please to ownerId].

coap-gateway/authorization/plgd.go
```diff
@@ -29,21 +29,21 @@ func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger, owne
...
-		Config:      config,
-		OAuth2:      &oauth2,
-		HTTPClient:  httpClient,
-		OwnerClaim:  ownerClaim,
-		UserInfoURL: oidcfg.UserInfoURL,
+		Config:     config,
+		OAuth2:     &oauth2,
+		HTTPClient: httpClient,
+		OwnerClaim: ownerClaim,
+		OpenID:     oidcfg,
```
ondrejtomcik, line 35: I don't think we need this configuration.

coap-gateway/config.yaml
```diff
@@ -60,24 +60,6 @@ clients:
         keyFile: "/secrets/private/cert.key"
         certFile: "/secrets/public/cert.crt"
         useSystemCAPool: false
```
ondrejtomcik, line 62: RA and RD shall be updated as well. We should drop service token completely.

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
 		return ctx, expire.IsValid()
 	}
 }
+
+type jwtClaims jwt.MapClaims
+
+func (s *Service) ValidateToken(ctx context.Context, token string) (jwtClaims, error) {
```
ondrejtomcik, line 33: It's GetClaims, ParseToken , but validate token would mean you verify if it's not expired, if the audience is correct, etc.

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
...
+	if err != nil {
+		return nil, err
+	}
+	return jwtClaims(m), nil
+}
+
+/// Get expiration time (exp) from user info map.
+/// It might not be set, in that case zero time and no error are returned.
+func (u jwtClaims) getExpirationTime() (time.Time, error) {
+	const expKey = "exp"
```
ondrejtomcik, line 46: Why are you doing it manually? JWT token parser, library which I think we used somewhere can verify if it's expired or not directly.

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
...
+		return time.Time{}, fmt.Errorf("invalid jwtClaims: invalid %v value type", expKey)
+	}
+	return pkgTime.Unix(int64(exp), 0), nil
+}
+
+/// Validate that ownerClaim is set and that it matches given user ID
+func (u jwtClaims) validateOwnerClaim(ocKey string, userID string) error {
+	v, ok := u[ocKey]
+	if !ok {
+		return fmt.Errorf("invalid jwtClaims: %v not set", ocKey)
```
ondrejtomcik: jwtClaims are not invalid. They are okay. Be more concrete in log messages please.
"configured owner claim %v is not present in the device token" where %v is value from configuration -> ownerClaim

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
...
+
+	exp, ok := v.(float64) // all integers are float64 in json
+	if !ok {
+		return time.Time{}, fmt.Errorf("invalid jwtClaims: invalid %v value type", expKey)
+	}
+	return pkgTime.Unix(int64(exp), 0), nil
+}
+
+/// Validate that ownerClaim is set and that it matches given user ID
+func (u jwtClaims) validateOwnerClaim(ocKey string, userID string) error {
```
ondrejtomcik: no ocKey . It's named ownerClaim only.

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
...
+	v, ok := u[ocKey]
+	if !ok {
+		return fmt.Errorf("invalid jwtClaims: %v not set", ocKey)
+	}
+	ownerClaim, ok := v.(string)
+	if !ok {
+		return fmt.Errorf("invalid jwtClaims: %v", "invalid ownerClaim value type")
```
ondrejtomcik: same here. "owner claim is present but it has an invalid type: %v"

coap-gateway/service/auth.go
```diff
@@ -68,3 +27,47 @@ func NewAuthInterceptor() Interceptor {
...
+	}
+	if ownerClaim != userID {
+		return fmt.Errorf("invalid ownerClaim: %v", userID)
```
ondrejtomcik: Again. owner claim is valid...

"owner identifier from the token '%v' doesn't match the one from the sign up message '%v'

docs/guide/deployment/coap-gateway.md
```diff
@@ -87,12 +87,6 @@ clients:
         caPool: "/data/certs/root_ca.crt"
         keyFile: "/data/certs/http.key"
         certFile: "/data/certs/http.crt"
-    oauth:
```
ondrejtomcik, line 90: Please document warning to this section: https://plgd.dev/architecture/component-overview/#cloud-registration that "Only JWT tokens are supported on the device."

ondrejtomcik (changes requested):

coap-gateway/authorization/plgd.go
jkralik: I need it to get JWKUrl

coap-gateway/config.yaml
jkralik: I don't want to make PR with 2000lines :)

coap-gateway/service/auth.go
jkralik: It's internal validate it (third party package) and if it's ok it's returns claims

coap-gateway/service/auth.go
jkralik: we need it to to parse - yes it is verify but I need to send it as response to device

coap-gateway/config.yaml
ondrejtomcik: Ok, but it will follow after this one, right ?

coap-gateway/service/auth.go
ondrejtomcik: Even internally, If I would want to use this method, I would probably use it wrongly.

coap-gateway/service/auth.go
jkralik: you are in package - you don't know who call you ... ` match the one from the sign up message '%v'`

coap-gateway/config.yaml
jkralik: cancelation I want to do as next ... I started this change because sign in returns bad expiresin to the device. Because auth0 doesn't support introspection of token ...

jkralik (commented):
done

pkg/security/jwt/validator.go
jkralik: fixed

coap-gateway/service/auth.go
jkralik: Validator provides `Validate` function which validates and parses the claims. I don't want to change it just for you ;D

coap-gateway/authorization/plgd.go
jkralik: Fixed

docs/guide/deployment/coap-gateway.md
jkralik: https://github.com/plgd-dev/plgd.dev/pull/15

ondrejtomcik (approved):

PR #390, opened 2021-08-09
Publish device Registered / Unregistered events to the EventBus

### LINKED ISSUE #392: Publish device Registered / Unregistered events to the EventBus
Our application using plgd is event-driven, subscribed directly to NATS events. We're able to find all resource-related events except device `REGISTERED` and `UNREGISTERED` events as they are defined [here](https://github.com/plgd-dev/cloud/blob/v2/grpc-gateway/pb/events.proto#L18).

Building a fully event-driven application on top of plgd. Currently, as 2 events are missing it's not possible.

We could pull information from the authorization server, but such a design doesn't meet our requirements.

jkralik:


authorization/pb/events.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package ocf.cloud.auth.pb;
+
+option go_package = "github.com/plgd-dev/cloud/authorization/pb;pb";
+
+// devices was registered to owner.
+message RegisteredDevices {
```
ondrejtomcik: The event described what has happened.
`DevicesRegistered`and `DevicesUnregistered`.

authorization/pb/events.proto
```diff
@@ -0,0 +1,30 @@
...
+    string owner = 1; // determined by clients.storage.ownerClaim.
+    repeated string device_ids = 2; // list of device ids added to owner.
+    int64 timestamp = 3; // unix timestamp in nanoseconds of creation event.
```
ondrejtomcik, line 16: I miss our AuditContext, which is helpful if somebody is building an audit log and you want to track who shared the device with this new owner.

authorization/pb/events.proto
```diff
@@ -0,0 +1,30 @@
...
+
+// devices was deregistered from owner.
+message UnregisteredDevices {
+    string owner = 1; // determined by clients.storage.ownerClaim
+    repeated string device_ids = 2; // list of device ids deleted from owner.
+    int64 timestamp = 3; // unix timestamp in nanoseconds of creation event.
+}
+
+// nats: owners.{owner}.>
+message Event {
```
ondrejtomcik, line 29: For resource-related events, we don't have such an envelope. Developers working with NATS won't like two different approaches. I would organize event types per subjects, as we do with RA.

jkralik: No. As a developer sometimes I want to be subscribed at  `owners.{owner}.>` to get all events. If I subscribe lower then I need to get the subject that provided the event. And I don't want to push the dependence to nats to the protocol.

ondrejtomcik: It's okay to organize it per as you defined -> `owners.{owner}.>`. My point is that `DevicesRegistered` shall be pushed to `owners.{owner}.deviceregistered` and `DevicesUnregistered` to `owners.{owner}.deviceunregistered`. This follows existing concepts you already introduced in RA.

Danielius1922 (approved):

Danielius1922 (changes requested):
Static analysis failed

ondrejtomcik (approved):

Danielius1922 (approved):

PR #408, opened 2021-08-21
resource-directory: lower cyclomatic complexity

Danielius1922:


resource-directory/service/subscriptions.go
```diff
@@ -449,29 +449,65 @@ func (s *Subscriptions) cancelSubscription(localSubscriptions *sync.Map, subscri
...
+
+	for _, sub := range subs {
+		err := s.Close(sub, nil)
+		if err != nil {
+			log.Errorf("cannot close subscription('%s'): %w", sub, err)
+		}
+	}
+}
+
+func (s *Subscriptions) getSubscriptionResponse(req *pb.SubscribeToEvents, localSubscriptions *sync.Map) (*pb.Event, bool) {
```
jkralik: this is weird - it will better when we split it to 2 functions `getSubscriptionResponse` and `getErrorSubscriptionResponse` and `getSubscriptionResponse` returns (*pb.Event, error) and when err != nil then `getErrorSubscriptionResponse` and send it.

jkralik (commented):
pls fix the comment

resource-directory/service/subscriptions.go
Danielius1922: How to propagate `SubscriptionId` from the first error case though?

```go
if err != nil {
  subRes.GetOperationProcessed().ErrorStatus.Code = pb.Event_OperationProcessed_ErrorStatus_ERROR
  subRes.GetOperationProcessed().ErrorStatus.Message = err.Error()
}
subRes.SubscriptionId = r.GetSubscriptionId()
```

I can only think of generating a custom error type and store it there.

jkralik (approved):

PR #410, opened 2021-08-23
coap-gateway: owner cache integration into service

Danielius1922:


coap-gateway/config.yaml
```diff
@@ -6,6 +6,7 @@ apis:
     address: "0.0.0.0:5684"
     externalAddress: ""
     maxMessageSize: 262144
+    cacheExpiration: 10m
```
ondrejtomcik: It's not descriptive enough. ownerCachceExpiration? Should it be really limited in the case of CoAP Gateway which should keep the cache open as long as the device is connected? There shouldn't be imho any time-based expiration.

jkralik: It's nice when you connect `1000 devices` at the same time so it will use data from the updated cache.

ondrejtomcik: Decrease to 1m

Danielius1922: Updated

resource-aggregate/cqrs/eventbus/nats/subscriber/subscriber.go
```diff
@@ -134,45 +135,57 @@ func WithGoPool(goroutinePoolGo eventbus.GoroutinePoolGoFunc) GoroutinePoolGoOpt
...
+		return nil, fmt.Errorf("cannot create cert manager: %w", err)
+	}
+	config.Options = append(config.Options, nats.Secure(certManager.GetTLSConfig()), nats.MaxReconnects(-1))
+	conn, err := nats.Connect(config.URL, config.Options...)
+	if err != nil {
+		certManager.Close()
+		return nil, fmt.Errorf("cannot create nats connection: %w", err)
+	}
+
+	s, err := NewWithNATS(conn, config.PendingLimits, logger, opts...)
```
jkralik: who close conn function? `NewWithNATS` I expected that the caller close conn. And with `New` I expected that will be close by subscribers. 
Can we rename `NewWithNATS` to just `New` and drop the old `New`?

resource-aggregate/cqrs/eventbus/nats/publisher/publisher.go
```diff
@@ -53,39 +54,49 @@ func WithMarshaler(dataMarshaler MarshalerFunc) MarshalerOpt {
...
 	}
 	p.AddCloseFunc(certManager.Close)
 	return p, nil
 }
 
-// NewPublisher creates a publisher.
-func newPublisher(url string, jetstream bool, eventMarshaler MarshalerFunc, options ...nats.Option) (*Publisher, error) {
-	conn, err := nats.Connect(url, options...)
+// Create publisher with existing NATS connection
+func NewWithNATS(conn *nats.Conn, jetstream bool, opts ...Option) (*Publisher, error) {
```
jkralik: can we rename `NewWithNATS` to `New` and drop the old `New`?

resource-aggregate/cqrs/eventbus/nats/subscriber/subscriber.go
Danielius1922: The current `subscriber.New` is used 14 times and `publisher.New` is used 33 times (mostly in tests though), so when I get rid of the original `New` then some convenience method equivalent would be useful.

Danielius1922: Solved by creating helper function for tests.

ondrejtomcik (approved):

jkralik (approved):

jkralik (approved):

PR #415, opened 2021-08-24
Jkralik/fix/projections

jkralik:


resource-directory/service/deviceMetadataProjection.go
```diff
@@ -130,7 +131,7 @@ func (p *deviceMetadataProjection) Handle(ctx context.Context, iter eventstore.I
 				return err
 			}
 			if err := p.data.HandleDeviceMetadataUpdatePending(ctx, &e); err != nil {
-				return err
+				goto nextEvent
```
Danielius1922: continue doesn't work for a for loop when called from an inner switch?

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -131,25 +131,19 @@ func (e *DeviceMetadataSnapshotTaken) Handle(ctx context.Context, iter eventstor
 			if err := eu.Unmarshal(&s); err != nil {
 				return status.Errorf(codes.Internal, "%v", err)
 			}
-			if err := e.HandleDeviceMetadataSnapshotTaken(ctx, &s); err != nil {
-				return err
-			}
+			_ = e.HandleDeviceMetadataSnapshotTaken(ctx, &s)
```
Danielius1922: Feels weird to just ignore all these handle errors, wouldn't at least logging then be useful?

Danielius1922 (approved):

resource-directory/service/deviceMetadataProjection.go
jkralik: fixed

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: Removed error returns for snapshot handlers.

ondrejtomcik (approved):

Danielius1922 (approved):

PR #443, opened 2021-09-02
OwnerCache in RA

Danielius1922:


jkralik (approved):

resource-aggregate/service/service.go
```diff
@@ -114,24 +113,24 @@ func NewService(ctx context.Context, config Config, logger log.Logger, eventStor
...
+		return nil, fmt.Errorf("cannot create nats client: %w", err)
+	}
+	grpcServer.AddCloseFunc(nats.Close)
+
+	ownerCache := clientAS.NewOwnerCache("sub", config.APIs.GRPC.OwnerCacheExpiration, nats.GetConn(), authClient, func(err error) {
+		log.Errorf("ownerCache error: %w", err)
+	})
+	grpcServer.AddCloseFunc(ownerCache.Close)
+
 	requestHandler := NewRequestHandler(config, eventStore, publisher, func(ctx context.Context, owner string, deviceIDs []string) ([]string, error) {
```
jkralik, line 117: maybe we can split into 2 functions. because we are don't need to always copy all devices for each action

jkralik: it is about O(n) vs O(ln(n))

resource-aggregate/service/service.go
```diff
@@ -114,24 +113,24 @@ func NewService(ctx context.Context, config Config, logger log.Logger, eventStor
```
Danielius1922, line 103: Get rid of `oauthMgr`, include part in configuration

jkralik (approved):

PR #446, opened 2021-09-03
Remove default annotations

Danielius1922:
Change default value of expiration in coap-gateway and
resource-aggregate for owner cache from 10m to 1m.

### DIFF
--- a/coap-gateway/service/config.go
+++ b/coap-gateway/service/config.go
@@ -59,9 +59,9 @@ func (c *APIsConfig) Validate() error {
 type COAPConfig struct {
 	Addr                     string                  `yaml:"address" json:"address"`
 	ExternalAddress          string                  `yaml:"externalAddress" json:"externalAddress"`
-	MaxMessageSize           int                     `yaml:"maxMessageSize" json:"maxMessageSize" default:"262144"`
-	OwnerCacheExpiration     time.Duration           `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration" default:"10m"`
-	GoroutineSocketHeartbeat time.Duration           `yaml:"goroutineSocketHeartbeat" json:"goroutineSocketHeartbeat" default:"4s"`
+	MaxMessageSize           int                     `yaml:"maxMessageSize" json:"maxMessageSize"`
+	OwnerCacheExpiration     time.Duration           `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration"`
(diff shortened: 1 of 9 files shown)


coap-gateway/service/config.go
```diff
@@ -60,7 +60,7 @@ type COAPConfig struct {
 	Addr                     string                  `yaml:"address" json:"address"`
 	ExternalAddress          string                  `yaml:"externalAddress" json:"externalAddress"`
 	MaxMessageSize           int                     `yaml:"maxMessageSize" json:"maxMessageSize" default:"262144"`
-	OwnerCacheExpiration     time.Duration           `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration" default:"10m"`
+	OwnerCacheExpiration     time.Duration           `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration" default:"1m"`
```
jkralik: default is not used by yaml parser - so you can remove it

resource-aggregate/service/config.go
```diff
@@ -44,7 +44,7 @@ func (c *APIsConfig) Validate() error {
 }
 
 type GRPCConfig struct {
-	OwnerCacheExpiration time.Duration `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration" default:"10m"`
+	OwnerCacheExpiration time.Duration `yaml:"ownerCacheExpiration" json:"ownerCacheExpiration" default:"1m"`
```
jkralik: default is not used by yaml parser - so you can remove it

Danielius1922: I see, I'll get rid of the defaults then.

Danielius1922: I've removed it from fields with `yaml` and `json` annotations, but kept it if `envconfig` was present (for example: 	https://github.com/plgd-dev/cloud/blob/v2/resource-directory/service/config.go#L63)

jkralik: It is not used anymore in any https://github.com/plgd-dev/cloud/blob/v2/<SERIVICE>/service/config.go.

Only https://github.com/plgd-dev/cloud/blob/v2/pkg/config pkg can contains env var for loading config path.

jkralik (commented):
If you need you can clean up all services.

resource-aggregate/service/config.go
Danielius1922: Removed from everything except c2c stuff, which can be removed only after the services are refactored.

jkralik (approved):

PR #448, opened 2021-09-03
Jkralik/feature/refactore grpc subs

### LINKED ISSUE #380: Support JWT tokens on devices

jkralik:


grpc-gateway/subscription/filterBitmask.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		res = append(res, val)
+	}
+	return res
+}
+
+func EventFilterToBitmask(f pb.SubscribeToEvents_CreateSubscription_Event) FilterBitmask {
+	bitmask := FilterBitmask(0)
+	switch f {
```
Danielius1922: This is a long switch, perhaps we could consider using array/map instead.

grpc-gateway/subscription/subscription.go
```diff
@@ -0,0 +1,792 @@
...
+		return
+	}
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if s.closed {
+		return
+	}
+	ev.CorrelationId = s.correlationID
+	ev.SubscriptionId = s.id
+	err = s.send(ev)
```
Danielius1922, line 300: Doing IO/communication with locked mutex is suboptimal. I think we can tolerate wring `unlock()` twice here (before line 287 and before `send`) instead of `defer`.

grpc-gateway/subscription/subscription.go
```diff
@@ -0,0 +1,792 @@
...
+func (s *Sub) Id() string {
+	return s.id
+}
+
+func (s *Sub) Init(ownerCache *client.OwnerCache) error {
+	owner, err := grpc.OwnerFromTokenMD(s.ctx, ownerCache.OwnerClaim())
+	if err != nil {
+		return grpc.ForwardFromError(codes.InvalidArgument, err)
+	}
+	s.mutex.Lock()
```
Danielius1922: This is a long mutex, that locks several functions, we should consider only copying necessary data under the mutex and then invoking functions (if possible).

grpc-gateway/service/subscribeToEvents.go
```diff
@@ -2,71 +2,208 @@ package service
...
 func (r *RequestHandler) SubscribeToEvents(srv pb.GrpcGateway_SubscribeToEventsServer) (errRet error) {
 	ctx, cancel := context.WithCancel(srv.Context())
 	defer cancel()
-	rd, err := r.resourceDirectoryClient.SubscribeToEvents(ctx)
-	if err != nil {
-		return log.LogAndReturnError(kitNetGrpc.ForwardErrorf(codes.Internal, "cannot subscribe to events: %v", err))
+
+	var sendMutex sync.Mutex
+	send := func(e *pb.Event) error {
+		sendMutex.Lock()
```
Danielius1922, line 152: I guess that `srv.Send` cannot be invoked from multiple goroutines concurrently.  Though locking whole functions instead of data can cause deadlocks/slower. 

We can keep it if it works, but I'm wondering what would be the _clean_ mechanism in Go without a `mutex`. An unbuffered channel and a single goroutine that would read events from this channel and invoke `srv.Send`...?

Danielius1922 (approved):

grpc-gateway/subscription/filterBitmask.go
jkralik: Fixed

grpc-gateway/subscription/subscription.go
jkralik: Fixed -> added goroutine with buffered channel

grpc-gateway/subscription/subscription.go
jkralik: Fixed

grpc-gateway/service/subscribeToEvents.go
jkralik: `srv.Send` is not goroutine safe so we need to use sync primitive and the `mutex` is cheaper than `channel + goroutine`.

grpc-gateway/service/config.go
```diff
@@ -30,7 +31,27 @@ func (c *Config) Validate() error {
...
+	SubscriptionBufferSize      int           `yaml:"subscriptionBufferSize" json:"subscriptionBufferSize"`
+	server.Config               `yaml:",inline" json:",inline"`
+}
+
+func (c *GRPCConfig) Validate() error {
+	if c.OwnerCacheExpiration <= 0 {
+		return fmt.Errorf("ownerCacheExpiration('%v')", c.OwnerCacheExpiration)
+	}
+	if c.SubscriptionCacheExpiration <= 0 {
+		return fmt.Errorf("subscriptionCacheExpiration('%v')", c.OwnerCacheExpiration)
```
Danielius1922: `c.OwnerCacheExpiration` -> `c.SubscriptionCacheExpiration`

coap-gateway/service/clientObserveHandler.go
```diff
@@ -71,92 +72,89 @@ func SendResourceContentToObserver(client *Client, resourceChanged *events.Resou
...
-				if _, err := client.cancelResourceSubscription(token, false); err != nil {
+	sub := subscription.New(req.Context, client.server.resourceSubscriber, client.server.rdClient, func(e *pb.Event) error {
+		switch {
+		case e.GetResourceUnpublished() != nil:
+			if !strings.Contains(e.GetResourceUnpublished().GetHrefs(), href) {
+				return nil
+			}
+			fallthrough
+		case e.GetDeviceUnregistered() != nil:
+			err := client.server.taskQueue.Submit(func() {
```
Danielius1922: lines 100-108 and 112-120 are the same, function/closure?

coap-gateway/service/clientObserveHandler.go
```diff
@@ -71,92 +72,89 @@ func SendResourceContentToObserver(client *Client, resourceChanged *events.Resou
...
+	}, &pb.SubscribeToEvents_CreateSubscription{
+		ResourceIdFilter:    []string{res.ToString()},
+		EventFilter:         []pb.SubscribeToEvents_CreateSubscription_Event{pb.SubscribeToEvents_CreateSubscription_RESOURCE_CHANGED, pb.SubscribeToEvents_CreateSubscription_UNREGISTERED, pb.SubscribeToEvents_CreateSubscription_RESOURCE_UNPUBLISHED},
+		IncludeCurrentState: true,
+	})
 	_, loaded := client.resourceSubscriptions.LoadOrStore(token, sub)
 	if loaded {
-		if _, err := sub.Cancel(); err != nil {
+		if err := sub.Close(); err != nil {
 			log.Errorf("failed to cancel resource /%v%v subscription: %w", deviceID, href, err)
```
Danielius1922: cancel -> close

coap-gateway/service/clientObserveHandler.go
```diff
@@ -71,92 +72,89 @@ func SendResourceContentToObserver(client *Client, resourceChanged *events.Resou
...
 		}
 		client.logAndWriteErrorResponse(fmt.Errorf("DeviceId: %v: cannot observe resource /%v%v: resource subscription with token %v already exist", authCtx.GetDeviceID(), deviceID, href, token), coapCodes.BadRequest, req.Token)
 		return
 	}
 
+	err = sub.Init(client.server.ownerCache)
+	if err != nil {
+		_, _ = client.resourceSubscriptions.PullOut(token)
+		if err := sub.Close(); err != nil {
+			log.Errorf("failed to cancel resource /%v%v subscription: %w", deviceID, href, err)
```
Danielius1922: cancel -> close

grpc-gateway/subscription/subscription.go
```diff
@@ -0,0 +1,840 @@
...
+func (s *Sub) Id() string {
+	return s.id
+}
+
+func (s *Sub) initOwnerSubscription(ownerCache *client.OwnerCache) ([]string, error) {
+	owner, err := grpc.OwnerFromTokenMD(s.ctx, ownerCache.OwnerClaim())
+	if err != nil {
+		return nil, grpc.ForwardFromError(codes.InvalidArgument, err)
+	}
+	s.mutex.Lock()
```
Danielius1922: What's being locked by `s.mutex` here? It doesn't seem necessary to lock the `ownerCache.Subscribe`, so can't this be moved after the `Subscribe` call?

grpc-gateway/service/config.go
jkralik: Fixed

coap-gateway/service/clientObserveHandler.go
jkralik: fixed

coap-gateway/service/clientObserveHandler.go
jkralik: fixed

coap-gateway/service/clientObserveHandler.go
jkralik: fixed

docs/guide/deployment/coap-gateway.md
```diff
@@ -137,6 +137,8 @@ CoAP API as specified in the [Open Connectivity Foundation - Device to Cloud Ser
 | `api.coap.externalAddress` | string | `External address including public domain/IP for coap client connection.` | `"coap-gw.example.com:5684"` |
 | `api.coap.maxMessageSize` | int | `Max message size which can be send/receive via coap. i.e. 256*1024 = 262144 bytes.` | `262144` |
 | `api.coap.ownerCacheExpiration` | string | `Time limit of how long to keep subscribed to device updates after last use of the given cache item.` | `1m` |
+| `api.grpc.subscriptionCacheExpiration` | string | `Time limit of how long to keep initialized events in subscription cache for duplicating events.` | `1m` |
```
ondrejtomcik: The time limit of how long to keep initialized events in the subscription cache is required for event deduplication.

ondrejtomcik: Isn't 1 minute too much? What exactly is the use-case here - when the event duplication might occur?

docs/guide/deployment/coap-gateway.md
```diff
@@ -137,6 +137,8 @@ CoAP API as specified in the [Open Connectivity Foundation - Device to Cloud Ser
...
+| `api.grpc.subscriptionBufferSize` | int | `The maximum buffer size for one events subscription.` | `1000` |
```
ondrejtomcik, line 141: for one events subscription? For one event - what does it mean, if the buffer size is set to 1000 events?

grpc-gateway/config.yaml
```diff
@@ -3,6 +3,9 @@ log:
 apis:
   grpc:
     address: "0.0.0.0:9100"
+    ownerCacheExpiration: 1m
```
ondrejtomcik, line 6: @rafajpet ping

grpc-gateway/config.yaml
```diff
@@ -36,6 +39,7 @@ apis:
           useSystemCAPool: false
 clients:
   authorizationServer:
+    ownerClaim: sub
```
ondrejtomcik, line 41: @rafajpet ping

grpc-gateway/pb/service.proto
```diff
@@ -99,7 +99,8 @@ service GrpcGateway {
     };
   }
 
-  // Subscribe to events
+  // When the client creates a subscription, with include_current_state=true it can send duplicate events.
```
ondrejtomcik, line 102: It's nice to have it here, but it belongs to swagger. That's where users would like to be informed about this. I asked in a different comment, but why duplication occurs during subscription? Is it due to the race condition which occurs when you subscribe + retrieve from RD, while during retrieval from RD some event might be already received + eventually also processed and returned by the RD?

docs/guide/deployment/coap-gateway.md
jkralik: This depends - if lot's of events comes eg registering 10000 devices with 42 resources then 1 minute is too small because it's needed to be bigger because registration of so much devices take sometime (5+ minutes)

docs/guide/deployment/coap-gateway.md
jkralik: for one subscription - there is a buffer (queue) where all events from the event bus are stored during initialization or processing events.

grpc-gateway/pb/service.proto
jkralik: yes correct

docs/guide/deployment/coap-gateway.md
ondrejtomcik: It's not a good design if it's not deterministic on our end. We cannot modify deployment now and then just because more or fewer devices are being registered. What you confirmed is that it's kept there for the time when you subscribe and do retrieve to RD. Why execution of these 2 operations cannot define the time of the "cache"? Why you cannot just subscribe, process events, call rpc call, merge and return to client merged?

grpc-gateway/pb/service.proto
ondrejtomcik: I don't like that client has to do deduplication in the websocket. Why you cannot get values, merge them with current subscription diff and send after that?

jkralik: Otherwise, we need to hold version per resource even type: for 10000 devices with 40 resources, it is approx 400000 hashes which holds version per subscription. cca 30MB RAM per subscription so it's not cheap and we need to dropped in one time.

ondrejtomcik: I don't mean deduplication during the whole time subscription exists. This wouldn't be a good idea. Only initial request when include_current_state=true is set. It's the same as I mentioned in a previous comment.

jkralik: Do you know the time? I don't know it because processing events from NATS can take more than initialization.

ondrejtomcik: It’s the time of response from RD. Not from nats?

jkralik: It's not guaranteed :), the time from RD and processing events from NATS are not the same. 
If time from NATS queue processing <= RD processing then it's OK
But how do you want to guarantee it when the time for NATS queue processing > RD processing? - this can cause duplication

ondrejtomcik: Nats doesn’t confirm successful subscription?

jkralik: no - and it doesn't matter - confirm successful subscription is about lose events , for deduplication it doesn't helps

grpc-gateway/config.yaml
ondrejtomcik: ignore. will be removed :)

docs/guide/deployment/coap-gateway.md
ondrejtomcik: As discussed, including current state will be deprecated and cache is no more required.

ondrejtomcik (approved):

PR #449, opened 2021-09-03
[ui] Pending commands page

kuby:


ondrejtomcik (approved):

http-gateway/web/src/containers/pending-commands/_pending-commands-list.js
```diff
@@ -0,0 +1,333 @@
+import { useEffect, useMemo, useState } from 'react'
```
oskarbauer, line 1: why are you naming the files with _ at the beginning, a new convention?

http-gateway/web/src/containers/pending-commands/constants.js
```diff
@@ -0,0 +1,59 @@
...
+  ACCEPTED: 'ACCEPTED',
+  ERROR: 'ERROR',
+  METHOD_NOT_ALLOWED: 'METHOD_NOT_ALLOWED',
+  CREATED: 'CREATED',
+  CANCELED: 'CANCELED',
+}
+
+export const PENDING_COMMANDS_LIST_REFRESH_INTERVAL_MS = 7000
+
+export const dateFormat = {
```
oskarbauer, line 44: maybe move these into an i18n file so they are easily found.

oskarbauer (approved):

http-gateway/web/src/containers/pending-commands/_pending-commands-list.js
kuby: Yes, kinda. Smaller parts are with _ and the containers/pages are without it.

http-gateway/web/src/containers/pending-commands/constants.js
kuby: Okay, once I start using dates on other places as well, I'll move it to a different location. For now, I would leave it in the constants, since its the closest place for it to be.

PR #456, opened 2021-09-10
coap-gateway: load deviceID from cert for mtls

### LINKED ISSUE #380: Support JWT tokens on devices

jkralik:


coap-gateway/service/signOff.go
```diff
@@ -109,26 +109,34 @@ func signOffHandler(req *mux.Message, client *Client) {
...
+
 	ctx = kitNetGrpc.CtxWithToken(ctx, signOffData.accessToken)
-	deviceIds := []string{signOffData.deviceID}
+	deviceIds := []string{deviceID}
 	respRA, err := client.server.raClient.DeleteDevices(ctx, &commands.DeleteDevicesRequest{
 		DeviceIds: deviceIds,
 	})
 	if err != nil {
-		logErrorAndCloseClient(fmt.Errorf("cannot handle sign off: %w", err), coapCodes.InternalServerError)
+		logErrorAndCloseClient(fmt.Errorf("cannot handle sign off: %w", err), coapCodes.Unauthorized)
```
Danielius1922: I think `DeleteDevices` can also fail because of some mongodb error, returning Unathorized doesn't seem correct in that case. Though `raClient.DeleteDevices` always returns `codes.Internal`, we should updated to `Unauthenticated` when user doesn't have access and then translate the code here to correct coap code.

Same with `AddDevice` call in signUp.

jkralik: Fixed

coap-gateway/coapconv/grpcErr2CoapCode.go
```diff
@@ -60,3 +61,11 @@ func GrpcCode2CoapCode(statusCode codes.Code, operation Operation) coapCodes.Cod
 	}
 	return coapCodes.InternalServerError
 }
+
+func GrpcErr2CoapCode(err error, operation Operation) coapCodes.Code {
```
Danielius1922, line 65: the `operation` parameter is never used in the body of the function

coap-gateway/service/resourceDiscovery.go
```diff
@@ -62,7 +60,7 @@ func makeDiscoveryResp(isTLSListener bool, serverAddr string, getResourceLinksCl
 			break
 		}
 		if err != nil {
-			return nil, status.Convert(err).Code(), fmt.Errorf("cannot create discovery response: %w", err)
+			return nil, fmt.Errorf("cannot create discovery response: %w", err)
```
Danielius1922, line 63: Does `GrpcErr2CoapCode` still work correctly when you wrap the original `err` with `fmt.Errorf`?

http-gateway/swagger.yaml
```diff
@@ -751,6 +751,9 @@ components:
           description: JWT claim identifying device owner.
           type: string
           default: 'sub'
+        jwtDeviceIdClaim:
+          description: JWT claim identifying device id. Empty means that JWT doesn't contains it.
```
Danielius1922: typo: JWT doesn't contains it -> JWT doesn't contain it

coap-gateway/config.yaml
```diff
@@ -52,6 +52,7 @@ clients:
         useSystemCAPool: false
   authorizationServer:
     ownerClaim: sub
+    deviceIdClaim: ""
```
ondrejtomcik: `deviceIdClaim` is retrieved and used by the coap gateway layer, which handles sign-in,sign-up and so on. Why is it part of the client->authorizationServer configuration? In my opinion it belongs to apis->coap->authorization

coap-gateway/service/auth.go
```diff
@@ -33,46 +31,39 @@ func NewAuthInterceptor() Interceptor {
...
 
-/// Get expiration time (exp) from user info map.
-/// It might not be set, in that case zero time and no error are returned.
-func (u jwtClaims) getExpirationTime() (time.Time, error) {
-	const expKey = "exp"
-	v, ok := u[expKey]
-	if !ok {
-		return time.Time{}, nil
+func (s *Service) VerifyDeviceID(tlsDeviceID string, claim pkgJwt.Claims) error {
+	if !s.config.APIs.COAP.TLS.Enabled {
```
ondrejtomcik, line 54: Could you please document this on the web? Developer documentation, advanced security. Document this simple condition there, how to enforce verification that the device token belongs to the device which does the sign-in.

coap-gateway/service/auth.go
```diff
@@ -33,46 +31,39 @@ func NewAuthInterceptor() Interceptor {
...
-	if !ok {
-		return time.Time{}, fmt.Errorf("expiration claim '%v' is present, but it has an invalid type '%T'", expKey, v)
+	if !s.config.APIs.COAP.TLS.Embedded.ClientCertificateRequired {
+		return nil
+	}
+	if s.config.Clients.AuthServer.DeviceIDClaim == "" {
+		return nil
+	}
+	jwtDeviceID := claim.DeviceID(s.config.Clients.AuthServer.DeviceIDClaim)
+	if jwtDeviceID != tlsDeviceID {
```
ondrejtomcik, line 66: Helpful would be to also log if the deviceId claim is not present. With your error we know that it doesn't match, but we can have one more error that it's not present in the token.

coap-gateway/service/auth.go
```diff
@@ -33,46 +31,39 @@ func NewAuthInterceptor() Interceptor {
...
+		return fmt.Errorf("deviceID('%v') in JWT doesn't match deviceID from token (%v)", jwtDeviceID, tlsDeviceID)
```
ondrejtomcik: Don't you use `:` before `%v` instead of `(%v)`?
`access token issued to the device ('%v') used by the different device ('%v')`

coap-gateway/service/auth.go
```diff
@@ -33,46 +31,39 @@ func NewAuthInterceptor() Interceptor {
...
-	v, ok := u[ownerClaim]
-	if !ok {
-		return fmt.Errorf("owner claim '%v' is not present", ownerClaim)
-	}
-	owner, ok := v.(string)
-	if !ok {
-		return fmt.Errorf("owner claim '%v' is present, but it has an invalid type '%T'", ownerClaim, v)
+func (s *Service) GetDeviceID(claim pkgJwt.Claims, tlsDeviceID, paramDeviceID string) string {
+	if s.config.Clients.AuthServer.DeviceIDClaim != "" {
+		return claim.DeviceID(s.config.Clients.AuthServer.DeviceIDClaim)
```
ondrejtomcik: What if the claim is not present in the token?

coap-gateway/service/auth.go
```diff
@@ -33,46 +31,39 @@ func NewAuthInterceptor() Interceptor {
...
-/// Validate that ownerClaim is set and that it matches given user ID
-func (u jwtClaims) validateOwnerClaim(ownerClaim string, userID string) error {
-	v, ok := u[ownerClaim]
-	if !ok {
-		return fmt.Errorf("owner claim '%v' is not present", ownerClaim)
-	}
-	owner, ok := v.(string)
-	if !ok {
-		return fmt.Errorf("owner claim '%v' is present, but it has an invalid type '%T'", ownerClaim, v)
+func (s *Service) GetDeviceID(claim pkgJwt.Claims, tlsDeviceID, paramDeviceID string) string {
```
ondrejtomcik: Why you don't pass here the client, which has authContext, tlsDeviceId? Or why is not the receiver type Client?

ondrejtomcik: There is another GetDeviceId on authContext. It's a bit misleading which one does what when to use which. Could you please make it more clear?

http-gateway/swagger.yaml
jkralik: Fixed

coap-gateway/coapconv/grpcErr2CoapCode.go
jkralik: Fixed

coap-gateway/service/client.go
```diff
@@ -885,7 +888,7 @@ func (client *Client) UpdateDeviceMetadata(ctx context.Context, event *events.De
 	return err
 }
 
-func (c *Client) ValidateToken(ctx context.Context, token string) (jwtClaims, error) {
+func (c *Client) ValidateToken(ctx context.Context, token string) (pkgJwt.Claims, error) {
```
ondrejtomcik, line 891: I miss tests validating expected scenario `when the device uses the token which doesn't belong to it` and `when a different device executes the code exchange - code belongs to a different device (signUp)` and other scenarios like deviceIdClaim not present in the token, ...

coap-gateway/service/resourceDiscovery.go
jkralik: Yes. It uses https://pkg.go.dev/errors#As which unwraps grpc error. If it is not a grpc error then `GrpcErr2CoapCode ` returns an internal error. (`codes.Unknown -> coapCodes.InternalServerError`)

coap-gateway/service/service.go
```diff
@@ -173,7 +215,33 @@ func New(ctx context.Context, config Config, logger log.Logger) (*Service, error
 			return nil, fmt.Errorf("cannot create tls cert manager: %w", err)
 		}
 		nats.AddCloseFunc(coapsTLS.Close)
-		l, err := net.NewTLSListener("tcp", config.APIs.COAP.Addr, coapsTLS.GetTLSConfig())
+		tlsCfgForClient := coapsTLS.GetTLSConfig()
+		var tlsCfg tls.Config
```
ondrejtomcik: Don't you want to add this new part to a different func ? This `New()` is already pretty huge and unreadable.

coap-gateway/service/signIn.go
```diff
@@ -258,55 +258,63 @@ func signInPostHandler(req *mux.Message, client *Client, signIn CoapSignInReq) {
...
-		logErrorAndCloseClient(fmt.Errorf("cannot handle sign in: %v", err), coapCodes.InternalServerError)
+		logErrorAndCloseClient(fmt.Errorf("cannot handle sign in: %w", err), coapCodes.InternalServerError)
 		return
 	}
 
-	if err := jwtClaims.validateOwnerClaim(client.server.config.Clients.AuthServer.OwnerClaim, signIn.UserID); err != nil {
-		logErrorAndCloseClient(fmt.Errorf("cannot handle sign in: %v", err), coapCodes.InternalServerError)
+	err = client.server.VerifyDeviceID(client.tlsDeviceID, jwtClaims)
+	if err != nil {
+		logErrorAndCloseClient(fmt.Errorf("cannot handle sign in: %w", err), coapCodes.Unauthorized)
```
ondrejtomcik, line 273: When unauthorized is during SignIn returned, the device resets it's coapcloudconf - removes at from the storage. Correct?

grpc-gateway/pb/cloudConfiguration.proto
```diff
@@ -11,12 +11,14 @@ message CloudConfigurationResponse {
     // provides a current time of server in nanoseconds.
     int64 current_time = 1;
     string device_onboarding_code_url  = 2;
-    string jwt_claim_owner_id = 3;
+    string jwt_owner_claim = 3;
+    string jwt_device_id_claim = 11;
```
ondrejtomcik: You can still renumber, we removed some stuff and v2 was not yet released.

grpc-gateway/pb/cloudConfiguration.proto
```diff
@@ -11,12 +11,14 @@ message CloudConfigurationResponse {
...
     string signing_server_address  = 4;
     string cloud_id = 5;
     string cloud_url  = 6;
     string cloud_authorization_provider = 7;
     string cloud_certificate_authorities = 8;
+    string device_authority = 10;
```
ondrejtomcik: There is nothing like device authority. There is only one server issuing tokens, and this one is called authority.
Please rename to `authorization_server`
Please update documentation of authority everywhere to:
```
Authority is the address of the token-issuing authentication server. Services will use this URI to find and retrieve the public key that can be used to validate the token’s signature.
```

ondrejtomcik (changes requested):

coap-gateway/service/auth.go
jkralik: No. It is obvious from error.

coap-gateway/config.yaml
jkralik: I understand but do we want to really JWT claims spread over the whole configuration? I expected that they are near. And still, device id is used for authorization in case mtls+deviceIdClaim

coap-gateway/service/signIn.go
jkralik: No. The Device tries to refresh the token and if the refresh token returns unauthorized then the device sets itself to failed state.

coap-gateway/service/auth.go
jkralik: `Don't you use : before %v instead of (%v)?` - it is a standard because you forward error from inner function

coap-gateway/service/client.go
jkralik: Added test

coap-gateway/service/service.go
jkralik: moved to separate func

grpc-gateway/pb/cloudConfiguration.proto
jkralik: fixed

grpc-gateway/pb/cloudConfiguration.proto
jkralik: fixed

coap-gateway/service/auth.go
jkralik: this is different but I change the name to `client.ResolveDeviceID()`

it is not cheap to evaluate per call deviceID so we need to resolve it and store it to authCtx.

coap-gateway/config.yaml
ondrejtomcik: > device id is used for authorization in case mtls+deviceIdClaim

yes, but it has nothing to do with the authorization server client. From the configuration perspective, which at the end represents the system, it looks like the grpc client which communicates with the authorization server needs to know the device claim key. That's not the case. It's the authorization part of the CoAP Gateway, API, which verifies it.

So exactly this: device id is used for authorization in case mtls+deviceIdClaim
is coap.api's responsibility.

coap-gateway/service/auth.go
ondrejtomcik: So are you able to identify that you misconfigured the auth0 and the claim is not there at all vs your client requesting the onboarding codes for the devices has a bug as it won't forward correct device id in the query param?

coap-gateway/config.yaml
jkralik: moved to apis.coap.authorization

coap-gateway/service/auth.go
jkralik: https://github.com/plgd-dev/cloud/blob/jkralik/feature/coap-gw-verify-deviceID/coap-gateway/service/auth.go#L52

ondrejtomcik (approved):

Danielius1922 (approved):

coap-gateway/service/auth.go
jkralik: https://github.com/plgd-dev/plgd.dev/pull/22

PR #461, opened 2021-09-14
c2c-gateway: update required scopes in API

### LINKED ISSUE #406: [cloud2cloud-gateway] Events APIs for registering subscriptions asks the client to have "w:subscriptions:* which is not matched with the spec and CTT requirement.
@ondrejtomcik @jkralik 

In the mail thread with OCF staff about "Why are CT5.3.1~CT5.3.3 test cases using only Retrieval Scope for Account Linking to get an access token ?", I have a lot of discussion about this. Finally the reason why I didn't know  the scope for subscription was changed to only retrieval scope and the implementation of plgd was not updated to the final spec.     

### Description
[cloud2cloud-gateway] Events APIs for registering subscriptions asks the client to have "w:subscriptions:* which is not matched with the final spec and CTT requirement. So I think it need to be fixed as soon as possible.

Danielius1922:
Update scopes to match specification.

### DIFF
--- a/cloud2cloud-gateway/service/service.go
+++ b/cloud2cloud-gateway/service/service.go
@@ -38,45 +38,43 @@ var authRules = map[string][]kitNetHttp.AuthArgs{
 		{
 			URI: regexp.MustCompile(`[\/]+api[\/]+v1[\/]+devices[\/]*$`),
 			Scopes: []*regexp.Regexp{
-				regexp.MustCompile(`r:deviceinformation:.*`),
+				regexp.MustCompile(`r:.*`),
 			},
 		},
 		{
 			URI: regexp.MustCompile(`[\/]+api[\/]+v1[\/]+devices[\/]*\?` + ContentQuery + `=` + ContentQueryBaseValue + `$`),
 			Scopes: []*regexp.Regexp{
-				regexp.MustCompile(`r:deviceinformation:.*`),
+				regexp.MustCompile(`r:.*`),
 			},
 		},
 		{
 			URI: regexp.MustCompile(`[\/]+api[\/]+v1[\/]+devices[\/]*\?` + ContentQuery + `=` + ContentQueryAllValue + `$`),
 			Scopes: []*regexp.Regexp{
-				regexp.MustCompile(`r:deviceinformation:.*`),
-				regexp.MustCompile(`r:resources:.*`),
+				regexp.MustCompile(`r:.*`),
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


cloud2cloud-gateway/service/service.go
```diff
@@ -38,97 +38,89 @@ var authRules = map[string][]kitNetHttp.AuthArgs{
...
 			},
 		},
 	},
 	http.MethodDelete: {
 		{
 			URI: regexp.MustCompile(`[\/]+api[\/]+v1[\/]+devices[\/]+subscriptions[\/]+[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}[\/]*$`),
 			Scopes: []*regexp.Regexp{
-				regexp.MustCompile(`r:deviceinformation:.*`),
-				regexp.MustCompile(`w:subscriptions:.*`),
+				regexp.MustCompile(`r:.*`),
```
Danielius1922, line 111: I'm not yet aware of what do these scopes mean, but the fact that `Delete` doesn't require `write` seems weird to me. Is it that OK?

jkralik: By spec https://openconnectivity.org/specs/OCF_Cloud_API_For_Cloud_Services_Specification_v2.2.4.pdf it is OK.

jkralik (approved):

cloud2cloud-gateway/service/service.go
Trishia: @Danielius1922 

See below answer of Mart Trayer about read scope of subscription POST/DELETE APIs from email thread discussion: 
> Greetings,
>
>You’re conflating the HTTP method used with what actually happens to the underlying data.
>
>A subscription for events is akin to an observe in CoAP (and other protocols), it doesn’t change any underlying data, it simply reports on changes triggered by others.  You don’t need to have update privileges (or scope) in order to be able to retrieve information/data.  The POST/DELETE is creating/removing the subscription, it is not changing the data that is being notified, hence “r:*” is perfectly acceptable and reasonable. 
>
>Best,
>mark.

Trishia (approved):

ondrejtomcik (approved):

cloud2cloud-gateway/service/service.go
Danielius1922: @Trishia, I see, thanks for the explanation.

PR #472, opened 2021-09-26
make scope cfg value optional in http gw

ondrejtomcik:


http-gateway/service/config.go
```diff
@@ -130,9 +130,6 @@ func (c *OAuthClientConfig) Validate() error {
 	if c.ClientID == "" {
 		return fmt.Errorf("clientID('%v')", c.ClientID)
 	}
-	if c.Audience == "" {
```
Danielius1922, line 133: Didn't you want to remove the scope verification?

jkralik, line 133: Commit message says about optional scopes but you remove audience. Is it correct?

jkralik (approved):

Danielius1922 (approved):

PR #473, opened 2021-09-27
resource-directory: refactoring

Danielius1922:


resource-directory/service/filterBitmask.go
```diff
@@ -44,13 +44,12 @@ func filterPendingCommandToBitmask(f pb.GetPendingCommandsRequest_Command) filte
 }
 
 func filterPendingsCommandsToBitmask(commandFilter []pb.GetPendingCommandsRequest_Command) filterBitmask {
```
jkralik: maybe you can use filterBitmask from https://github.com/plgd-dev/cloud/blob/v2/grpc-gateway/subscription/filterBitmask.go

jkralik:
Did you update the c2c-gw docs for change RD to grpc-gw?

resource-directory/service/filterBitmask.go
Danielius1922: Yes, it's the same. Switched to use the grpc-gw/subscription code.

Danielius1922:
> Did you update the c2c-gw docs for change RD to grpc-gw?

Docs updated.

authorization/client/ownerCache.go
```diff
@@ -109,6 +109,9 @@ func (d *ownerSubject) syncDevicesLocked(ctx context.Context, owner string, cach
 	return added, removed, nil
 }
 
+// ErrFunc reports errors
+type ErrFunc func(err error)
```
jkralik: change `type ErrFunc func(err error)` to alias `type ErrFunc = func(err error)` we don't need to be strict

jkralik (approved):

PR #475, opened 2021-09-28
Rename authority service to identity service

Danielius1922:

PR #479, opened 2021-09-30
pkg/oauth2: update provider and token types

Danielius1922:
Remove OwnerClaim from provider and Owner from token. OwnerClaim
value is available from stored configuration; owner is set in
access token.

### DIFF
--- a/coap-gateway/service/refreshToken.go
+++ b/coap-gateway/service/refreshToken.go
@@ -148,7 +148,7 @@ func refreshTokenPostHandler(req *mux.Message, client *Client) {
 		return
 	}
 
-	owner := token.Owner
+	owner := claim.Owner(client.server.config.APIs.COAP.Authorization.OwnerClaim)
 	if owner == "" {
 		owner = refreshToken.UserID
 	}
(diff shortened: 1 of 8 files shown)


coap-gateway/service/signUp.go
```diff
@@ -116,6 +116,13 @@ func signUpPostHandler(r *mux.Message, client *Client) {
 		return
 	}
 
+	// client.server.config.APIs.COAP.Authorization.OwnerClaim
+	owner := claim.Owner("sub")
```
jkralik: Why sub?

coap-gateway/service/refreshToken.go
```diff
@@ -148,7 +148,8 @@ func refreshTokenPostHandler(req *mux.Message, client *Client) {
 		return
 	}
 
-	owner := token.Owner
+	// client.server.config.APIs.COAP.Authorization.OwnerClaim
+	owner := claim.Owner("sub")
```
jkralik: Why sub

coap-gateway/service/signUp.go
Danielius1922: The `client.server.config.APIs.COAP.Authorization.OwnerClaim` path is not valid in v2 yet, once #475 is merged then I can use it.

ondrejtomcik (approved):

jkralik (approved):

PR #489, opened 2021-10-01
resource-aggregate: fill audit context to event resourceLinksSnapshotTaken from the last modifier

jkralik:


ondrejtomcik (approved):

resource-aggregate/commands/utils.go
```diff
@@ -44,7 +44,7 @@ func NewResourceID(deviceID, href string) *ResourceId {
 }
 
 func (r *Resource) IsObservable() bool {
-	return r.GetPolicies() != nil && r.GetPolicies().GetBitFlags()&2 != 0
+	return r.GetPolicy() != nil && r.GetPolicy().GetBitFlags()&2 != 0
```
Danielius1922: That `& 2` is a bit obscure, isn't there a constant we can use instead of the magic number?

Danielius1922 (approved):

Danielius1922 (approved):

PR #491, opened 2021-10-01
Use consistent naming for identity service

Danielius1922:
* package, binary and configuration sections are named "identity"
* in logs and documentation the full name "identity service" is used
* usage of "identity server" was removed


jkralik (approved):
@ondrejtomcik I think it is OK. What do you think?

ondrejtomcik (approved):

jkralik (approved):

ondrejtomcik (approved):

PR #500, opened 2021-10-05
Domain renamed to authority, it must include protocol as well

kuby:
@ondrejtomcik @jkralik 

New web config structure

```json
{
  "authority": "",
  "httpGatewayAddress": "",
  "webOAuthClient": {
    "clientID": "",
    "audience": "",
    "scopes": []
  },
  "deviceOAuthClient": {
    "clientID": "",
    "audience": "",
    "scopes": [],
    "providerName": ""
  }
}
```

### DIFF
--- a/http-gateway/web/src/index.js
+++ b/http-gateway/web/src/index.js
@@ -18,11 +18,11 @@ fetch('/web_configuration.json')
     const audience = config?.webOAuthClient?.audience
     const scopes = config?.webOAuthClient?.scopes?.join?.(',') || ''
     const httpGatewayAddress = config.httpGatewayAddress
-    const domain = config.domain
+    const authority = config.authority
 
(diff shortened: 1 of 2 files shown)


http-gateway/web/src/index.js
```diff
@@ -46,7 +46,7 @@ fetch('/web_configuration.json')
         <Provider store={store}>
           <IntlProvider>
             <Auth0Provider
-              domain={domain}
+              domain={authority}
```
jkralik, line 49: is it ok? The previous property domain doesn't contain `https://`, now it is part of the property authority.

kuby: I have tested it and it works with the protocol as well, but please test it too if you can, to make sure it works for 100%.

ondrejtomcik (approved):

PR #507, opened 2021-10-08
grpc-gateway: share subscriptions among clients

jkralik:


grpc-gateway/subscription/convertToSubjects_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}{
+		{
+			name: "empty",
+		},
+		{
+			name: "all",
+			args: args{
+				bitmask: FilterBitmaskMax,
+			},
+			want: []string{"plgd.owners.1b4db7eb-4057-5ddf-91e0-36dec72071f5.>"},
```
Danielius1922: What are these hard-coded `1b4db7eb-4057-5ddf-91e0-36dec72071f5`, `e1407479-3136-56c0-9908-bb02fb0339e2`, ... values? 

They seem to be related to `owner`. Creating function to convert owner to id and then using it instead of the hardcoded values would help readability.

Danielius1922 (approved):

grpc-gateway/subscription/convertToSubjects_test.go
jkralik: Fixed

ondrejtomcik (approved):

PR #509, opened 2021-10-12
rename cloud-configuration to hub-configuration

jkralik:


ondrejtomcik (approved):

resource-directory/test/service.go
```diff
@@ -26,11 +26,11 @@ func MakeConfig(t *testing.T) service.Config {
...
 
-	cfg.ExposedCloudConfiguration.CAPool = config.CA_POOL
-	cfg.ExposedCloudConfiguration.AuthorizationServer = "https://" + config.OAUTH_SERVER_HOST
-	cfg.ExposedCloudConfiguration.CloudID = config.CloudID()
-	cfg.ExposedCloudConfiguration.CloudURL = config.GW_HOST
-	cfg.ExposedCloudConfiguration.OwnerClaim = config.OWNER_CLAIM
+	cfg.ExposedHubConfiguration.CAPool = config.CA_POOL
+	cfg.ExposedHubConfiguration.AuthorizationServer = "https://" + config.OAUTH_SERVER_HOST
+	cfg.ExposedHubConfiguration.CloudID = config.CloudID()
+	cfg.ExposedHubConfiguration.CloudURL = config.GW_HOST
```
Danielius1922: Maybe we could drop the `Cloud` prefix for `CloudID` and `CloudURL` and keep only `ID` and `URL` as field names here?

Danielius1922 (approved):

resource-directory/test/service.go
ondrejtomcik: Good point. id is ok. But url, not sure. Endpoint? Address? Uri? Url? No idea. 🙂 And it's only coap endpoint. Right?

jkralik: CoapGateway is better, what do you think?

PR #510, opened 2021-10-12
test-cloud-server: build and publish test-cloud-server image

jkralik:


Danielius1922 (approved):

Danielius1922 (approved):

.github/workflows/publishTestCloudServer.yml
```diff
@@ -0,0 +1,57 @@
...
+          tags: |
+            type=schedule
+            type=ref,event=branch
+            type=ref,event=tag
+            type=ref,event=pr
+      
+      - name: Build hub-build image
+        run: make hub-build
+
+      - name: Build and push test-cloud-server docker image
```
ondrejtomcik, line 51: it took me 10 seconds to understand why our image is named cloud-server. Should it really be named by example application located in the different repository? what about test-bundle ?

jkralik: This test is independent of the bundle. It is binary which contains all tests from grpc-gateway/service. So I don't want to move it to bundle.

ondrejtomcik: So do you want to stick with bame cloud-server?

jkralik: It's a testing tool for an iotivity cloud server so yes.

ondrejtomcik (approved):

PR #512, opened 2021-10-13
Jkralik/fix/publishing messages

jkralik:


identity-store/service/addDevice.go
```diff
@@ -40,7 +40,8 @@ func (s *Service) publishDevicesRegistered(ctx context.Context, owner, userID st
 		return err
 	}
 
-	err = s.publisher.Flush(ctx)
+	// timeout si driven by flusherTimeout.
+	err = s.publisher.Flush(context.Background())
```
Danielius1922, line 44: Not entirely related, but `PublishData` ignores the `ctx` argument, so it can be removed and then `ctx` parameter can be now removed from `publishDevicesRegistered` and publishDevicesUnregistered` methods.

Danielius1922 (approved):

Danielius1922 (approved):

ondrejtomcik (approved):

PR #519, opened 2021-10-18
coap-gateway: don't allow refresh token for not registered device

jkralik:


coap-gateway/service/refreshToken_test.go
```diff
@@ -63,6 +64,10 @@ func TestRefreshTokenHandlerWithRetry(t *testing.T) {
 	defer func() {
 		_ = co.Close()
 	}()
+	testSignUp(t, CertIdentity, co)
+	defer func() {
```
Danielius1922: Already defered on lines 64-66.

jkralik: Fixed

Danielius1922 (approved):

PR #521, opened 2021-10-18
Helm chart improve docs

rafajpet:


charts/plgd-hub/values.yaml
```diff
@@ -331,6 +354,7 @@ resourcedirectory:
   apis:
```
jkralik, line 354: instead of `https://github.com/plgd-dev/hub/tree/main/resource-directory`  use `https://plgd.dev/configuration/resource-directory/#yaml-configuration`

pls apply the same pattern for all links

ondrejtomcik (approved):

PR #523, opened 2021-10-19
github/action: publish images to ghcr.io with arm64 platform

jkralik:


.github/workflows/test.yml
```diff
@@ -3,8 +3,13 @@
 name: Test
 
 # Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the master branch
+# events but only any branch
```
Danielius1922: not entirely correct, push is only on main and PR is on any branch

.github/workflows/publishDockerImagesGhcr.yml
```diff
@@ -0,0 +1,702 @@
...
+    - completed
+  push:
+    tags:
+    - '*'
+
+env:
+  REGISTRY: ghcr.io
+  VNEXT_BRANCH: main
+
+jobs:
```
Danielius1922, line 24: This will be a pain to maintain, it's basically the same job copied 12 times for each service. Surely GH workflow offer some sort of way of creating a job template or something for such a use case. Maybe [strategy](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategy) ?

Danielius1922 (approved):

.github/workflows/test.yml
jkralik: fixed

.github/workflows/publishDockerImagesGhcr.yml
jkralik: 👍 fixed via strategy

Danielius1922 (approved):

ondrejtomcik (approved):

PR #526, opened 2021-10-22
fix/resource-aggregate: repeated create and delete operations doesn't send the same set of events

Danielius1922:


resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -315,19 +293,13 @@ func (e *ResourceStateSnapshotTaken) HandleEventResourceChanged(ctx context.Cont
...
-		}
-	}
-	if index < 0 {
-		return status.Errorf(codes.InvalidArgument, "cannot find resource delete pending event with correlationId('%v')", deleted.GetAuditContext().GetCorrelationId())
-	}
+	e.LatestResourceChange = nil
+	e.ResourceCreatePendings = nil
+	e.ResourceRetrievePendings = nil
+	e.ResourceDeletePendings = nil
+	e.ResourceUpdatePendings = nil
```
jkralik: we can set it to only for the positive status (OK, ACCEPT,DELETED).

jkralik (approved):

PR #537, opened 2021-10-29
test: add mocked coap-gateway

Danielius1922:


Danielius1922:
The mocked coap-gateway useful for testing resource observation implementation, but I had to disable tests since they will succeed only after https://github.com/iotivity/iotivity-lite/pull/147 is merged.

jkralik (approved):

PR #549, opened 2021-11-01
Update go modules

Danielius1922:


jkralik (approved):

Danielius1922:
@jkralik doesn't something use a `git clone` command (maybe some docker image)? Because with a submodule we need to use `--recursive` option, I'll update the tutorial page in https://plgd.dev/quickstart/contribute/ , but is there something else that needs to be updated?

jkralik:
> @jkralik doesn't something use a `git clone` command (maybe some docker image)? Because with a submodule we need to use `--recursive` option, I'll update the tutorial page in https://plgd.dev/quickstart/contribute/ , but is there something else that needs to be updated?

No, we don't directly use `git clone` in any script.

Danielius1922:
> No, we don't directly use `git clone` in any script.

Should be fine then, it's only necessary for developers who want to run 'make proto/generate' since some of our .proto files include code from googleapis.

PR #559, opened 2021-11-05
[cloud2cloud-connector] fix oauth error in account linking logic.

Trishia:
@ondrejtomcik 
I pushed fixed codes as I mentioned in issue https://github.com/plgd-dev/hub/issues/532. Review and then merge plz if you have any problem.

### DIFF
--- a/cloud2cloud-connector/service/addLinkedAccount.go
+++ b/cloud2cloud-connector/service/addLinkedAccount.go
@@ -55,12 +55,22 @@ func (rh *RequestHandler) handleOAuth(w http.ResponseWriter, r *http.Request, li
 	if loaded {
 		return http.StatusInternalServerError, fmt.Errorf("cannot store key - collision")
 	}
-	oauthCfg := linkedCloud.OAuth
-	if oauthCfg.RedirectURL == "" {
-		oauthCfg.RedirectURL = rh.provider.Config.RedirectURL
+
+	if !linkedAccount.Data.HasOrigin() {
+		url := rh.provider.Config.AuthCodeURL(t)
+		http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+		return http.StatusOK, nil
 	}
-	url := oauthCfg.AuthCodeURL(t)
-	http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+
+	if !linkedAccount.Data.HasTarget() {
+		oauthCfg := linkedCloud.OAuth
+		if oauthCfg.RedirectURL == "" {
+			oauthCfg.RedirectURL = rh.provider.Config.RedirectURL
+		}
+		url := oauthCfg.AuthCodeURL(t)
+		http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+		return http.StatusOK, nil
+	}
 	return http.StatusOK, nil
 }
 
--- a/pkg/security/oauth2/plgd.go
+++ b/pkg/security/oauth2/plgd.go
@@ -20,6 +20,7 @@ func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger) (*Pl
 	if err != nil {
 		return nil, err
 	}
+	config.ClientSecret = string(clientSecret)
 
 	httpClient, err := client.New(config.HTTP, logger)
 	if err != nil {
@@ -30,7 +31,9 @@ func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger) (*Pl
 		return nil, err
 	}
 
-	oauth2 := config.ToOAuth2(oidcfg.AuthURL, oidcfg.TokenURL, string(clientSecret))
+	config.AuthURL = oidcfg.AuthURL
+	config.TokenURL = oidcfg.TokenURL
+	oauth2 := config.ToOAuth2(config.AuthURL, config.TokenURL, config.ClientSecret)
 
 	return &PlgdProvider{
 		Config:     config,
--- a/pkg/security/oauth2/plgd.go
+++ b/pkg/security/oauth2/plgd.go
@@ -33,11 +33,8 @@ func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger) (*Pl
 
 	config.AuthURL = oidcfg.AuthURL
 	config.TokenURL = oidcfg.TokenURL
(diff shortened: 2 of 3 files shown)


Trishia:
@ondrejtomcik @jkralik @Danielius1922 
Account linking was successfully completed with this commit but it seems pulling devices was not working still now.  I didn't see any message for "pulling devices for ... " in cloud2cloud-connector logs.

ondrejtomcik:
We will look into this MR @Trishia as soon as we will get into testing of C2C Connector. Currently we're on the C2C Gateway.

Trishia:
@ondrejtomcik 
When will you finish reviewing this merge request ?

ondrejtomcik:
Sometime in december. We are not working on c2c connector.
But what I don't understand is why you changed the clientSecret in the config. This shouldn't be changed. It's loaded only from the file, no other option won't be allowed within the container.

Trishia:
> Sometime in december. We are not working on c2c connector. But what I don't understand is why you changed the clientSecret in the config. This shouldn't be changed. It's loaded only from the file, no other option won't be allowed within the container.

As I said before in the gitter chat, client secret loaded from client secret file has made an issue related to oauth exchange. I don't know exact reason why this happen but the client secret data from file that is made by echo command (i.e. $ echo "D1tiNC6oFRLuXbRX8txliM7qILk56BJEnQWDHm9VvS00bwyML1Ol0NplXgXnGTN2" > clientSecret.dat) seem to be different from client secret config even if I checked the data via log print in the containers. 

Anyway, I will rollback clientSecret in the oauth config. Review and merge plz if it has no problem.

pkg/security/oauth2/plgd.go
```diff
@@ -12,14 +12,21 @@ import (
...
+	}
+	if config.ResponseType == "" {
+		config.ResponseType = "code"
+	}
 
 	clientSecret, err := file.Load(config.ClientSecretFile, make([]byte, 4096))
 	if err != nil {
 		return nil, err
 	}
+	config.ClientSecret = string(clientSecret)
```
Danielius1922, line 23: @jkralik could you check this PR please, without this the CTT Origin tests don't work. 

Seems OK to me, but fields `ClientSecret`, `AuthURL` and `TokenURL` seems are part of config, but also part of oauth2 (line 42) and now they have the same value. Maybe there's a way to do it better and also make the code in `addLinkedAccount.go` work, but I do not know how.

cloud2cloud-connector/service/addLinkedAccount.go
```diff
@@ -55,12 +55,22 @@ func (rh *RequestHandler) handleOAuth(w http.ResponseWriter, r *http.Request, li
 	if loaded {
 		return http.StatusInternalServerError, fmt.Errorf("cannot store key - collision")
 	}
-	oauthCfg := linkedCloud.OAuth
-	if oauthCfg.RedirectURL == "" {
-		oauthCfg.RedirectURL = rh.provider.Config.RedirectURL
+
+	if !linkedAccount.Data.HasOrigin() {
+		url := rh.provider.Config.AuthCodeURL(t)
```
jkralik, line 60: pls, replace `rh.provider.Config.AuthCodeURL(t)` with `rh.provider.OAuth2.AuthCodeURL(t)`

pkg/security/oauth2/plgd.go
jkralik: @Danielius1922 
lines 29 and 40,41,42 are not needed to be modified/added if you apply the comment in addLinkedAccount.go. So could you change it and test if it works?

jkralik (approved):

pkg/security/oauth2/plgd.go
jkralik: We can add call AuthCodeURL to provider config. Instead of `rh.provider.Config.AuthCodeURL(t)` we can call `rh.provider.AuthCodeURL(t)`

jkralik (approved):

pkg/security/oauth2/plgd.go
```diff
@@ -12,14 +12,21 @@ import (
 
 // NewPlgdProvider creates OAuth client
 func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger) (*PlgdProvider, error) {
-	config.ResponseMode = "query"
-	config.AccessType = "offline"
-	config.ResponseType = "code"
+	if config.ResponseMode == "" {
```
ondrejtomcik: @Trishia why you need to make this values configurable? None of these should be change from it's default values, so why to make it configurable?

pkg/security/oauth2/plgd.go
Danielius1922: I'm going to get rid of the whole `OAuth2` substruct as we talked about, it'll make things cleaner.

jkralik (commented):
Waiting for @Trishia response.

pkg/security/oauth2/plgd.go
Trishia: @ondrejtomcik  oauth struct has these values so I add default check which is safer source code for flexibility if developers want to change these configs.

Trishia:
> Waiting for @Trishia response.

@jkralik What kind of response do you wait ? Review changed codes or Merge this PR ?

pkg/security/oauth2/plgd.go
ondrejtomcik: But he shouldn't, that's the point. Response type should be always code, and it shouldn't be updated. so the question is if you have a use-case when you need to change these values

Trishia: If you think it should not be changed, rollback to original codes you want.

jkralik:
> > Waiting for @Trishia response.
> 
> @jkralik What kind of response do you wait ? Review changed codes or Merge this PR ?

@Trishia I mean the question from @ondrejtomcik :

[@Trishia why you need to make this values configurable? None of these should be change from it's default values, so why to make it configurable?](https://github.com/plgd-dev/hub/pull/559#discussion_r782931895)

pkg/security/oauth2/plgd.go
Danielius1922: Guys told me that I should keep those values hard-coded, so I'll revert the code and we can merge this.

jkralik (approved):

Danielius1922 (approved):

PR #562, opened 2021-11-12
publish plgd/cert-tool as docker image

jkralik:


tools/cert-tool/cmd/main.go
```diff
@@ -0,0 +1,154 @@
...
+		os.Exit(2)
+	}
+
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var cert []byte
+	switch {
```
Danielius1922, line 127: SonarCloud might complain that this is too complex of a function

Danielius1922 (approved):

tools/cert-tool/cmd/main.go
jkralik: fixed

PR #563, opened 2021-11-12
coap-gw: rework subscriptions

Danielius1922:


coap-gateway/coapconv/coapconv.go
```diff
@@ -237,6 +251,56 @@ func NewNotifyResourceChangedRequest(resourceID *commands.ResourceId, connection
 	}
 }
 
+func NewNotifyResourceChangedRequests(deviceID, connectionID string, req *pool.Message) ([]*commands.NotifyResourceChangedRequest, error) {
```
jkralik: rename function like `NewNotifyResourceChangedRequestsFromBatchResourceDiscovery`

coap-gateway/coapconv/coapconv.go
```diff
@@ -237,6 +251,56 @@ func NewNotifyResourceChangedRequest(resourceID *commands.ResourceId, connection
...
+		if len(resource.Content) == 2 {
+			var v map[interface{}]interface{}
+			if err := cbor.Decode(resource.Content, &v); err == nil && len(v) == 0 {
+				return true
+			}
+		}
+		return false
+	}
+
+	var requests []*commands.NotifyResourceChangedRequest
```
jkralik: preallocate array

coap-gateway/service/observation/resourcesObserver.go
```diff
@@ -0,0 +1,338 @@
...
+			initialized = true
+			observable := true
+			if _, errObs := msg.Observe(); errObs != nil {
+				log.Debugf("href: %v not observable err: %v", obsRes.Href(), errObs)
+				observable = false
+			}
+			if !observable {
+				return
+			}
+		}
```
jkralik: is this check needed -> we always want to send data via `o.callbacks.OnObserveResource` whether it is observable/not observable.

coap-gateway/service/observation/resourcesObserver.go
```diff
@@ -0,0 +1,338 @@
...
+//   - If the resource is observable then the connection to COAP-GW (coapConn) is used to
+//   register for observations from COAP-GW. Observation notifications are handled by the
+//   onObserveResource callback.
+//   - If the resource is not observable then a GET request is sent to COAP-GW to receive
+//   the content of the resource and the response is handled by the onGetResourceContent
+//   callback.
+type resourcesObserver struct {
+	lock      sync.Mutex
+	deviceID  string
+	resources map[int64]*observedResource // [instanceID]
```
jkralik: i'm thinking to change it to array -> unpublish happens very rarely and mostly all resources. array consumes less memory than map.

coap-gateway/service/observation/resourcesObserver.go
```diff
@@ -0,0 +1,338 @@
...
+	if r != nil {
+		return r.PopObservation()
+	}
+	return nil
+}
+
+func (o *resourcesObserver) removeResourceLocked(instanceID int64) {
+	delete(o.resources, instanceID)
+}
+
```
jkralik: do we need to have one line function ?

coap-gateway/service/resourceDirectory.go
```diff
@@ -61,46 +68,92 @@ func validatePublish(w wkRd) error {
...
-		return
+		return wkRd{}, fmt.Errorf("invalid publish request received from device %v: %w", deviceID, err)
 	}
 
 	w = fixTimeToLive(w)
 	for i, link := range w.Links {
 		w.Links[i].DeviceID = w.DeviceID
 		w.Links[i].Href = fixHref(link.Href)
-		w.Links[i].InstanceID = getInstanceID(link.Href)
+		w.Links[i].InstanceID = resource.GetInstanceID(link.Href)
```
jkralik, line 89: could you reset all other properties to empty values ? the response will be much smaller.

Danielius1922: You mean on `w.Links[i]`? But the data from it is used to generate `commands.PublishResourceLinksRequest`, which has a `Resources` field and that takes all the data from the link. Are you sure the data is not needed? If I follow our code correctly the resources are then stored in `ResourceLinksSnapshotTaken` struct in RA. 

However, it seems that RA only uses `DeviceID`, `Href` and `ResourceTypes` fields, so perhaps we can indeed simplify it.

coap-gateway/service/observation/resourcesObserver.go
Danielius1922: Yeah, I wanted to ask what to do here. `Observe` sends `Get` for non-observable resource. For non-observable `/oic/res` I want this to do nothing, no notification should be send and the observation as a whole will fall back to observing of published resources. And for published resources I should know whether the resource is observable or not. Only for observable resources will `coapConn.Observe` be called, non-observable call `coapConn.Get`. 

So if the detection of observability is reliable then this will work correctly. But if we want to be defensive then we can do it that for batch `/oic/res` it will do this and in other cases it will call `callbacks.OnObserveResource` (or maybe it should call `callbacks.OnGetResourceContent`? Since for non-observable resource it gets a `Get` response.)

coap-gateway/service/observation/resourcesObserver.go
Danielius1922: changed to sorted slice

coap-gateway/service/observation/resourcesObserver.go
Danielius1922: changed so this happens only for `/oic/res`

coap-gateway/service/resourceDirectory.go
```diff
@@ -110,7 +163,18 @@ func resourceDirectoryPublishHandler(req *mux.Message, client *Client) {
...
-	out, err := encode(w)
+
+	type wkRdResponse struct {
+		Links schema.ResourceLinks `json:"links,omitempty"`
+	}
+	var resp wkRdResponse
+	resp.Links = make(schema.ResourceLinks, len(w.Links))
+	for i, link := range w.Links {
+		resp.Links[i].Href = link.Href
+		resp.Links[i].InstanceID = link.InstanceID
```
jkralik: why you don't send deviceID and anchor...
image

Danielius1922: Only Href and IstanceID are used by iotivity-lite, so that's the baseline that must be included. I wasn't sure about anything else. 

I'm not sure what does the spec tell me. It says that the payload should have the same information as the original UPDATE request. Doesn't that mean that we should include everything in response (as was done originally)?

Danielius1922: Reverted to original after SM discussion.

jkralik (approved):

jkralik (approved):

PR #564, opened 2021-11-15
certificate-authority: don't sign ceritficate which contains hubId in…

jkralik:
… common name


ondrejtomcik:
Please update also plgd-dev/www

ondrejtomcik (approved):

PR #569, opened 2021-11-24
introduce consent screen and custom scopes in mock oauth server

ondrejtomcik:


test/oauth-server/service/token_test.go
```diff
@@ -62,6 +62,40 @@ func TestRequestHandler_getDeviceToken(t *testing.T) {
 	require.NotEmpty(t, accessToken[service.TokenDeviceID])
 }
 
+func TestRequestHandler_getTokenWithDefaultScopes(t *testing.T) {
```
jkralik: pls rename it to `TestRequestHandlerGetTokenWithDefaultScopes`

test/oauth-server/service/token_test.go
```diff
@@ -62,6 +62,40 @@ func TestRequestHandler_getDeviceToken(t *testing.T) {
...
+	require.Equal(t, token["scope"], "openid profile email")
+	validator := jwt.NewValidator(fmt.Sprintf("https://%s%s", config.OAUTH_SERVER_HOST, uri.JWKs), &tls.Config{
+		InsecureSkipVerify: true,
+	})
+	accessToken, err := validator.Parse(token["access_token"])
+	require.NoError(t, err)
+	require.Equal(t, "openid profile email", accessToken[service.TokenScopeKey])
+}
+
+func TestRequestHandler_getTokenWithCuscomScopes(t *testing.T) {
```
jkralik: pls rename it to `TestRequestHandlerGetTokenWithCuscomScopes`

test/oauth-server/uri/uri.go
```diff
@@ -16,10 +17,11 @@ const (
...
-	JWKs                = "/.well-known/jwks.json"
-	OpenIDConfiguration = "/.well-known/openid-configuration"
-	LogOut              = "/v2/logout"
+	Token                = "/oauth/token"
+	Authorize            = "/authorize"
+	UserInfo             = Authorize + "/userinfo"
+	JWKs                 = "/.well-known/jwks.json"
+	OpenIDConfiguration  = "/.well-known/openid-configuration"
+	LogOut               = "/v2/logout"
+	UnauthorizedResponse = "/authorize/unauthorized"
```
jkralik: Authorize + "/unauthorized"

jkralik:
The test fails because you pass the array as a joined string with "," so the scope is not an array as is expected by parser scopes:
https://github.com/plgd-dev/hub/blob/main/pkg/security/jwt/claims.go#L72

test/oauth-server/service/token.go
```diff
@@ -45,7 +45,7 @@ func makeAccessToken(clientID, host, deviceID string, issuedAt, expires time.Tim
 			return nil, fmt.Errorf("failed to set %v: %w", jwt.ExpirationKey, err)
 		}
 	}
-	if err := token.Set(TokenScopeKey, []string{"openid", "r:deviceinformation:*", "r:resources:*", "w:resources:*", "w:subscriptions:*"}); err != nil {
+	if err := token.Set(TokenScopeKey, scopes); err != nil {
```
jkralik, line 50: replace `TokenScopeKey` with https://github.com/plgd-dev/hub/blob/main/pkg/security/jwt/claims.go#L16

jkralik: maybe you need to convert it as an array if it contains contained scope with "," separator

ondrejtomcik: scopes is of type string, already in authorize.go defined as space separated scope list.

jkralik: ok so you need to split to array by " "

ondrejtomcik: No, because then you return comma separated items and that's wrong. I tested it. Returned should be items separated by comma. Until now it was wrong - separated by comma :) So it's a fix :)

Danielius1922 (approved):

jkralik (approved):

coap-gateway/service/refreshToken_test.go
```diff
@@ -68,7 +68,7 @@ func TestRefreshTokenHandlerWithRetry(t *testing.T) {
 
 	testRefreshToken := testEl{
 		name: "Retry",
-		in:   input{coapCodes.POST, `{"di": "` + CertIdentity + `", "uid":"` + AuthorizationUserId + `", "refreshtoken":"123"}`, nil},
+		in:   input{coapCodes.POST, `{"di": "` + CertIdentity + `", "uid":"` + AuthorizationUserId + `", "refreshtoken":"refresh-token"}`, nil},
```
Danielius1922, line 71: Does this mean that we always get "refresh-token" as RefreshToken regardless of the input?

test/oauth-server/service/config.go
```diff
@@ -26,10 +26,17 @@ const AllowedGrantType_PASSWORD AllowedGrantType = "password"
...
+	ClientSecret                    string        `yaml:"secret"`
+	AuthorizationCodeLifetime       time.Duration `yaml:"authorizationCodeLifetime"`
+	AccessTokenLifetime             time.Duration `yaml:"accessTokenLifetime"`
+	CodeRestrictionLifetime         time.Duration `yaml:"codeRestrictionLifetime"`
+	RefreshTokenRestrictionLifetime time.Duration `yaml:"refreshTokenRestrictionLifetime"`
+	ConsentScreenEnabled            bool          `yaml:"consentScreenEnabled"`
+	RequireIssuedAuthorizationCode  bool          `yaml:"requireIssuedAuthorizationCode"`
+	SupportedScope                  []string      `yaml:"supportedScope"`
+	SupportedResponseType           string        `yaml:"supportedResponseType"`
+	SupportedRedirectURI            string        `yaml:"supportedRecirectUri"`
```
Danielius1922: typo: `yaml:"supportedRecirectUri` -> `yaml:"supportedRedirectUri`

coap-gateway/service/refreshToken_test.go
ondrejtomcik: Regardless of what kind of input?
Yes. A refresh token with the value `"refresh-token"` is returned every time. A random value is generated only if you set `refreshTokenRestrictionLifetime` to other than the default value.

Danielius1922: This input:
```
input{coapCodes.POST, `{"di": "` + CertIdentity + `", "uid":"` + AuthorizationUserId + `", refreshtoken":"refreshtoken"}
```
I guess it previously automatically returned as RefreshToken what was in the `refreshtoken` value in the input, but now it returns `refresh-token` (unless `refreshTokenRestrictionLifetime` is set). The test is still valid, so this is OK, I guess I'm just confused about how does our mock oauth server work, it's beginning to be complicated. :D

Danielius1922:
Regarding `SonarCloud` analysis: I usually ignore low coverage if it's caused by negative code paths (ie error handling). However, since mock oauth is an integral part of most of our tests I would keep the coverage analysis enabled and try to keep coverage in the usual 70-80% realm (if possible), so we know it behaves as it should.

ondrejtomcik:
@Danielius1922 okay. I will do my homework 🙂

coap-gateway/service/refreshToken_test.go
ondrejtomcik: Hm, I am not aware of this functionality - set in the request to mock OAuth Server value of the refresh token which it then returns.

jkralik (approved):

ondrejtomcik:
@Danielius1922 better coverage? :)

jkralik:
Please update properly MR https://github.com/plgd-dev/www/pull/35

coap-gateway/service/refreshToken_test.go
ondrejtomcik: Ok, I checked it again. My understanding is, that this test simulates token refresh executed by the device. That means, that the CoAP Gateway tries to use this refresh token to get a new access token. Is my understanding correct?

My modification would then make sense because now the OAuth Server doesn't accept random refresh tokens unless you configure `refreshTokenRestrictionLifetime`. If this value is set to default, the refresh token received by the OAuth Server has to match `"refresh-token"`.

jkralik:
@ondrejtomcik  Did you look at smells (2 critical)?

Danielius1922:
> @Danielius1922 better coverage? :)

Yop, great coverage! Good job.

test/oauth-server/service/token_test.go
```diff
@@ -62,17 +63,229 @@ func TestRequestHandler_getDeviceToken(t *testing.T) {
...
+}
+
+func TestGetRequestHandlerGetTokenWithInvalidClient(t *testing.T) {
+	webTearDown := test.SetUp(t)
+	defer webTearDown()
+
+	getTokenUsingGet(t, "", false, http.StatusBadRequest)
+}
+
+// func TestRequestHandlerGetTokenWithFormContentType(t *testing.T) {
```
Danielius1922: remove?

test/oauth-server/test/test.go
```diff
@@ -45,28 +52,61 @@ func MakeConfig(t *testing.T) service.Config {
...
-			CodeRestrictionLifetime:   0,
+			ID:                              ClientTestC2C,
+			AuthorizationCodeLifetime:       time.Minute * 10,
+			AccessTokenLifetime:             0,
+			CodeRestrictionLifetime:         0,
+			RefreshTokenRestrictionLifetime: 0,
+			ConsentScreenEnabled:            true,
+		},
+		{
+			ID:                              ClientTestC2C,
```
Danielius1922: This clientID is here twice, is that OK?

Danielius1922 (approved):

PR #600, opened 2021-12-06
coap-gateway: fix unauthorized access test

jkralik:


coap-gateway/service/signIn_test.go
```diff
@@ -101,8 +101,12 @@ func TestSignInDeviceSubscriptionHandler(t *testing.T) {
 	require.True(t, cancelCtx.Err() == context.Canceled)
 
 	co1 := testCoapDial(t, testCfg.GW_HOST, "")
-	_, code := runSignIn(t, CertIdentity, signUpResp, co1)
-	require.Equal(t, coapCodes.Unauthorized, code)
+	resp, err := doSignIn(t, CertIdentity, signUpResp, co1)
```
Danielius1922, line 104: Previously `runSignIn` called `co.ReleaseMessage(resp)`, this is now skipped by calling `doSignIn` directly. Is that OK?

jkralik: Yes, it's ok. the message just will be not stored in the cache for reuse.

Danielius1922 (approved):

PR #604, opened 2021-12-08
Jkralik/fix/tls buffer conn

jkralik:


coap-gateway/service/publishedResources.go
```diff
@@ -14,9 +14,11 @@ type publishedResources struct {
 }
 
 func (p *publishedResources) Add(href ...string) {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-	p.resources = p.resources.Insert(href...)
+	/*
```
Danielius1922: This probably doesn't belong in this PR.

jkralik: fixed

Danielius1922:
The patch is not needed in `Dockerfile.test` file?

jkralik:
> The patch is not needed in `Dockerfile.test` file?

Good point. Fixed

Danielius1922 (approved):

PR #638, opened 2022-01-18
proto: use snake_case for variable names

Danielius1922:


ondrejtomcik (approved):

jkralik:
I think you need to update this too: https://github.com/plgd-dev/hub/blob/main/http-gateway/swagger.yaml#L1558

jkralik (changes requested):
My previous comment

Danielius1922:
> I think you need to update this too: https://github.com/plgd-dev/hub/blob/main/http-gateway/swagger.yaml#L1558

I wouldn't want to change that, all those keys in swagger are camel case, but I think everything is alright. This is the newly generated code:
```
TimeToLive int64 `protobuf:"varint,3,opt,name=time_to_live,json=timeToLive,proto3" json:"time_to_live,omitempty"` // command
```
I think it uses the `json` value inside the protobuf tag, so the value for the API didn't change. I can set the value from our GUI if I recall correctly, right? So I'm gonna check if everything is in order.

Danielius1922:
I've locally added logs to `grpc-gateway` and the TimeToLive value propagates correctly when I create/update/delete resource. And `http-gateway` uses camel case, these are my http requests:

```
PUT https://192.168.1.44/api/v1/devices/6e49b0e6-87fe-4405-4832-8be93e30be28/resources/switches?timeToLive=100000000&resourceInterface=oic.if.create
PUT https://192.168.1.44/api/v1/devices/6e49b0e6-87fe-4405-4832-8be93e30be28/resources/switches/1?timeToLive=200000000&resourceInterface=oic.if.a
DELETE https://192.168.1.44/api/v1/devices/6e49b0e6-87fe-4405-4832-8be93e30be28/resource-links/switches/1?timeToLive=500000000
```

So everything should work, this only modified internal naming used by `grpc-gateway`.

jkralik (approved):

PR #640, opened 2022-01-19
grpc-gateway: add update test case

Danielius1922:
Add test case to verify resource data consistency after device name change.
This is a test case for iotivility-lite issue #40 oic/d piid issue with notification (https://github.com/iotivity/iotivity-lite/issues/40).


jkralik:
Could you move tests to the `grpc-gateway/service`. Because only this package is used in iotivity-lite test container.

Danielius1922:
> Could you move tests to the `grpc-gateway/service`. Because only this package is used in iotivity-lite test container.

Service test added.

jkralik (approved):

PR #646, opened 2022-01-28
coap-gateway: allows to use client credentials for device provider

jkralik:


pkg/security/oauth2/plgd.go
```diff
@@ -33,58 +40,23 @@ func NewPlgdProvider(ctx context.Context, config Config, logger log.Logger) (*Pl
 
 	config.AuthURL = oidcfg.AuthURL
 	config.TokenURL = oidcfg.TokenURL
+	var provider provider
+	if config.GrantType == "clientCredentials" {
```
Danielius1922: A constant for `"clientCredentials"`?

Danielius1922 (approved):

Danielius1922:
[This PR](https://github.com/plgd-dev/hub/pull/648) should fix the failing test. I'll rebase this PR once I merge the test fix, it should pass all the tests then.

Danielius1922 (approved):

coap-gateway/config.yaml
```diff
@@ -49,6 +50,7 @@ apis:
           audience: ""
           redirectURL: "cloud.plgd.mobile://login-callback"
           authority: ""
+          grantType: ""
```
ondrejtomcik, line 53: did you updated www documentation @jkralik ?

pkg/security/oauth2/oauth/oauth.go
```diff
@@ -57,6 +64,13 @@ func (c *Config) Validate() error {
 	if c.ClientSecretFile == "" {
 		return fmt.Errorf("clientSecretFile('%v')", c.ClientSecretFile)
 	}
+	switch c.GrantType {
+	case ClientCredentials, AuthorizationCode:
```
ondrejtomcik, line 68: shouldn't we validate that ownerId is set if at least one client creds provider is configured on the coap-gateway? Reason is that if it comes to this combination, no ownerId set - sub is used, and client credential is used, super-device is created - it can query all devices. Can it?

PR #648, opened 2022-01-31
grpc-gateway: fix TestUpdateConfigurationName test

Danielius1922:
Add a delay after updating the device name so the subsequent GET sees the updated value.

### DIFF
--- a/grpc-gateway/client/updateResource_test.go
+++ b/grpc-gateway/client/updateResource_test.go
@@ -153,12 +153,17 @@ func TestUpdateConfigurationName(t *testing.T) {
 
 	startData := getData(deviceID)
 
-	name := "update simulator"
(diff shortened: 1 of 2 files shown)


grpc-gateway/client/updateResource_test.go
```diff
@@ -153,12 +153,17 @@ func TestUpdateConfigurationName(t *testing.T) {
...
 
-	name := "update simulator"
-	err := c.UpdateResource(ctx, deviceID, configuration.ResourceURI, map[string]interface{}{"n": name}, nil)
-	require.NoError(t, err)
+	updateName := func(name string) {
+		err := c.UpdateResource(ctx, deviceID, configuration.ResourceURI, map[string]interface{}{
+			"n": name,
+		}, nil)
+		require.NoError(t, err)
+		time.Sleep(time.Millisecond * 200)
```
jkralik, line 161: Maybe increase value to 1s

jkralik (approved):

PR #662, opened 2022-02-07
coap-gateway: avoid shuffled resource change events from the same resource

jkralik:


ondrejtomcik (approved):

grpc-gateway/service/updateResource_test.go
```diff
@@ -272,3 +273,69 @@ func TestRequestHandlerGetAfterUpdateResource(t *testing.T) {
...
+	defer func() {
+		_ = conn.Close()
+	}()
+	c := pb.NewGrpcGatewayClient(conn)
+
+	resources := test.GetAllBackendResourceLinks()
+	_, shutdownDevSim := test.OnboardDevSim(ctx, t, c, deviceID, config.GW_HOST, resources)
+	defer shutdownDevSim()
+
+	for i := 0; i < 10; i++ {
```
Danielius1922: I would increase the iterations to at least a 100, 10 might not be enough for a race condition to occur.

coap-gateway/service/client.go
```diff
@@ -167,7 +168,7 @@ func (client *Client) onObserveResource(ctx context.Context, deviceID, href stri
 		return fmt.Errorf("cannot handle resource observation: %w", err)
 	}
 	notification.Hijack()
-	err := client.server.taskQueue.Submit(func() {
+	err := client.server.taskQueue.SubmitForOneWorker(resource.GetInstanceID(deviceID+href), func() {
```
Danielius1922, line 171: Small optimization: we could calculate the `instanceID` value when observation is created and sent it to the callback as a parameter instead of always calculating it here in the callback.

jkralik: `resource.GetInstanceID` is very cheap function. So I don't want to optimize.

grpc-gateway/service/updateResource_test.go
jkralik: increased to 50

pkg/sync/task/queue/singleWorkerQueue.go
```diff
@@ -0,0 +1,107 @@
...
+	q.deleted = true
+}
+
+func (q *oneWorkerQueue) pop() func() {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+	if q.queue.Len() > 0 {
+		return q.queue.Remove(q.queue.Front()).(func())
+	}
+	q.freeLocked()
```
Danielius1922, line 44: This could be suboptimal for resources in heavy use, it will keep removing and adding the key to the `q.oneworkerQueue` map. But there's probably no need to optimize it now, we should see it in the profiler if we find a problematic use case.

Danielius1922 (approved):

grpc-gateway/service/updateResource_test.go
```diff
@@ -272,3 +273,70 @@ func TestRequestHandlerGetAfterUpdateResource(t *testing.T) {
...
+	c := pb.NewGrpcGatewayClient(conn)
+
+	resources := test.GetAllBackendResourceLinks()
+	_, shutdownDevSim := test.OnboardDevSim(ctx, t, c, deviceID, config.GW_HOST, resources)
+	defer shutdownDevSim()
+
+	for i := 0; i < numIteration; i++ {
+		t.Logf("TestRequestHandlerMultipleUpdateResource:run %v\n", i)
+		lightHref := test.TestResourceLightInstanceHref("1")
+		for i := 1; i >= 0; i-- {
```
Danielius1922: maybe a different indexing variable would be more readable, it's weird to see that both outer and inner use `i`.

jkralik: fixed

PR #669, opened 2022-02-18
coap-gateway: validate provide.grantType against ownerClaim

jkralik:


Danielius1922 (approved):

coap-gateway/service/config.go
```diff
@@ -83,6 +84,9 @@ func (c *AuthorizationConfig) Validate() error {
 	}
 	duplicitProviderNames := make(map[string]bool)
 	for i := 0; i < len(c.Providers); i++ {
+		if c.Providers[i].GrantType == oauth.ClientCredentials && c.OwnerClaim == "sub" {
+			return fmt.Errorf("providers[%v].grantType('%v') - %w", i, c.Providers[i].GrantType, fmt.Errorf("invalid combination with apis.coap.authorization.ownerClaim('%v')", c.OwnerClaim))
```
ondrejtomcik: Why error? In my opinion it's warn and ignore this provider / don't load it.

ondrejtomcik: combination of ownerClaim set to sub is not compatible if at least one authorization provider uses grant type ClientCredentials

ondrejtomcik (approved):

PR #687, opened 2022-03-27
Rafajpet/chart improvements

rafajpet:


certificate-authority/config.yaml
```diff
@@ -27,7 +27,7 @@ apis:
       certFile: "/secrets/private/cert.crt"
       clientCertificateRequired: true
     authorization:
-      ownerClaim: "sub"
```
ondrejtomcik: Why did you remove this from example `config.yaml`? Please revert.

ondrejtomcik: I would rather remove this here: https://github.com/plgd-dev/hub/blob/main/charts/plgd-hub/values.yaml#L436

charts/plgd-hub/templates/http-gateway/api-ingress.yaml
```diff
@@ -5,44 +5,31 @@
...
-        backend:
-          service:
-            name: {{ $fullname }}
-            port:
-              number: {{ $port }}
   - host: {{ include "plgd-hub.httpgateway.apiDomain" . | quote }}
     http:
       paths:
-      {{- range  .Values.httpgateway.ingress.paths }}
+      {{- range  .Values.httpgateway.ingress.api.paths }}
```
ondrejtomcik, line 32: Some customers are hosting API paths on the root domain and skip adding `api.` prefix by setting `grpcgateway.domain`, `httpgateway.apidomain` and `certificateauthority.domain` on their own to overwrite this automated prefix. Are you sure this won't be broken? Can you test such a deployment, where you set these properties to same value as global.domain and verify everything is accessible on the root domain?

coap-gateway/config.yaml
```diff
@@ -28,7 +28,7 @@ apis:
       certFile: "/secrets/public/cert.crt"
       clientCertificateRequired: true
     authorization:
-      ownerClaim: "sub"
```
ondrejtomcik: Please don't remove this from example config.yaml

charts/plgd-hub/templates/http-gateway/api-ingress.yaml
rafajpet: There is only change that value for `httpgateway.ingress.paths` was setup to `httpgateway.ingress.api.paths`. The same for ui part. Logic how ingress handle Path is still the same

ondrejtomcik (approved):

PR #700, opened 2022-04-06
resource-aggregate: generate events for each resource in unpublished resource links

jkralik:


Danielius1922 (approved):

grpc-gateway/pb/service.swagger.json
```diff
@@ -1108,6 +1109,9 @@
         },
         "deviceMetadataUpdated": {
           "$ref": "#/definitions/pbDeviceMetadataUpdated"
+        },
+        "resourceStateSnapshotTaken": {
+          "$ref": "#/definitions/pbResourceStateSnapshotTaken"
```
ondrejtomcik, line 1114: http gw swagger need to be updated as well

resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -536,78 +532,123 @@ func (e *ResourceStateSnapshotTaken) ConfirmResourceCreateCommand(ctx context.Co
...
 		return nil, status.Errorf(codes.NotFound, "cannot find commands with correlationID(%v)", req.GetCorrelationIdFilter())
 	}
 	return events, nil
 }
 
+func (e *ResourceStateSnapshotTaken) isEmpty() bool {
+	return e.GetLatestResourceChange() == nil && len(e.GetResourceCreatePendings()) == 0 && len(e.GetResourceDeletePendings()) == 0 && len(e.GetResourceRetrievePendings()) == 0 && len(e.GetResourceUpdatePendings()) == 0
+}
+
+func (e *ResourceStateSnapshotTaken) UnpublishResource(ctx context.Context, userID string, resourceID *commands.ResourceId, commandMetadata *commands.CommandMetadata, deletedContent *commands.Content, newVersion uint64) ([]eventstore.Event, error) {
```
ondrejtomcik: This function is not unpublishing resources. Please name it by describing what it does, not when it is used.

resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -712,6 +753,13 @@ func (e *ResourceStateSnapshotTaken) HandleCommand(ctx context.Context, cmd aggr
 		}
 		return []eventstore.Event{&rc}, nil
 	case *commands.ConfirmResourceDeleteRequest:
+		if req.GetCommandMetadata() == nil {
+			return nil, status.Errorf(codes.InvalidArgument, errInvalidCommandMetadata)
+		}
+		if req.GetStatus() == commands.Status_OK || req.GetStatus() == commands.Status_ACCEPTED {
+			// deleted can comes first than unpublish so we call same handler
```
ondrejtomcik: Deleted event might be received before ... better formulation please

resource-aggregate/service/grpcApi.go
```diff
@@ -155,6 +155,9 @@ func (r RequestHandler) UnpublishResourceLinks(ctx context.Context, request *com
 		log.Errorf("cannot publish resource links unpublished events: %v", err)
 	}
 	auditContext := commands.NewAuditContext(owner, "")
+
+	r.UnpublishResources(ctx, request, owner, events)
```
ondrejtomcik: This from flow doesn't make sense. You unpublished resource links on L148. What are you doing here when you are already creating a response? Why another request handler? It's not readable. Mainly because you call it unpublish resources, what has nothing to do with resource state. Call it for example CleanupResourceState. Including go file name, functions, everything.

resource-aggregate/test/service.go
```diff
@@ -26,7 +26,7 @@ func MakeConfig(t *testing.T) service.Config {
 
 	cfg.Clients.Eventstore.Connection.MongoDB = config.MakeEventsStoreMongoDBConfig()
 	cfg.Clients.Eventstore.ConcurrencyExceptionMaxRetry = 8
-	cfg.Clients.Eventstore.SnapshotThreshold = 16
+	cfg.Clients.Eventstore.SnapshotThreshold = 32
```
ondrejtomcik: Why did you increased it?

resource-aggregate/service/grpcApi.go
ondrejtomcik: Before the device signs off, it will unpublish all resources. Am I correct?
If that's true, this will lead to a performance bottleneck as we have to go through every published resource and cancel its pending requests and create a snapshot. Do we want to really do it? What if some device starts to unpublish all links before it signs off?

resource-aggregate/test/service.go
jkralik: grpc-gateway/service/TestCreateAndDeleteResource test doesn't expect the snapshot event during the test. So I moved the setup to the test.

resource-aggregate/service/grpcApi.go
jkralik: Fixed

resource-aggregate/events/resourceStateSnapshotTaken.go
jkralik: fixed

resource-aggregate/events/resourceStateSnapshotTaken.go
jkralik: removed

resource-aggregate/service/grpcApi.go
jkralik: Yes, we want to try to do it. It means that all clients that wait for pending commands will get cancelled and the cache will not contain the old resource.

grpc-gateway/service/createAndDeleteResource_test.go
```diff
@@ -198,9 +199,18 @@ func deleteSwitchResourceExpectedEvents(t *testing.T, deviceID, subID, correlati
 		},
 	}
 
+	snaphost := &pb.Event{
```
Danielius1922, line 202: snaphost -> snapshot

PR #701, opened 2022-04-07
coap-gatewaty: refactore COAP handlers for simple logging

jkralik:


coap-gateway/service/client.go
```diff
@@ -294,12 +295,11 @@ func (client *Client) onObserveResource(ctx context.Context, deviceID, href stri
 }
 
 // Close closes coap connection
-func (client *Client) Close() error {
+func (client *Client) Close() {
 	err := client.coapConn.Close()
 	if err != nil {
-		return fmt.Errorf("cannot close client: %w", err)
+		client.Errorf("cannot close client: %w", err)
```
Danielius1922, line 301: This simplifies the code, but we lose the context of the call. Won't it be a problem that we no longer identify the `Close` call exactly by the error message?

jkralik: This error is not very interesting -> because just informs you that a close connection fails which happened only when the connection was already closed by a peer.

Danielius1922 (approved):

PR #706, opened 2022-04-13
Rafajpet/helm config improvements

rafajpet:


charts/plgd-hub/README.md
```diff
@@ -64,6 +64,7 @@ global:
 | certificateauthority.config.fileName | string | `"service.yaml"` | File name for config file |
 | certificateauthority.config.mountPath | string | `"/config"` | Mount path |
 | certificateauthority.config.volume | string | `"config"` | Config file volume name |
+| certificateauthority.customPodPorts | object | `{}` | Custom certificate-authority container ports definition. In case it's not defined, default definition is used |
```
ondrejtomcik: Why won't you use `certificateauthority.service.ports` for this? Is there a reason to have a pod port defined even though you don't have it exposed through the service?

rafajpet: I try to figure out scenario where you would like to forward traffic from service itself not directly to main container within pod but to different container, let's say auth sidecar. You don't want to break api for ingress/other service consume this service so you don't want to change port for service but within POD you would like to switch port for sidecar with original container port.

ondrejtomcik: Do you have such a scenario now or is it just a theoretical expectation? Change only what is needed.

rafajpet: Add custom authorization logic for grpc-gateway, certificate-authority, http-gateway in form of sidecar for those services

ondrejtomcik: But we don't have custom authorization logic. Do we?

rafajpet: We would like to use it for kistler IoT platform. Special permission logic for subset of plgd-hub services

ondrejtomcik:
Is this the correct annotation for the HTTP gateway? 
https://github.com/plgd-dev/hub/blob/rafajpet/helm-config-improvements/charts/plgd-hub/values.yaml#L1086

ondrejtomcik (approved):

PR #716, opened 2022-04-25
Enable additional linters in golansci-lint

Danielius1922:


cloud2cloud-connector/events/subscription.go
```diff
@@ -18,7 +18,7 @@ const (
 	EventType_DevicesUnregistered EventType = "devices_unregistered"
 
 	// among all
-	EventType_SubscriptionCanceled EventType = "subscription_cancelled"
+	EventType_SubscriptionCanceled EventType = "subscription_canceled"
```
jkralik: this breaks spec  - is defined by `subscription_cancelled` OCF spec

Danielius1922: British vs American English, but I think I can set the locale in misspell.

Danielius1922: Reverted

jkralik (approved):

PR #725, opened 2022-05-06
implement otel to http servers

jkralik:


http-gateway/service/httpApi.go
```diff
@@ -112,7 +112,7 @@ func resourceEventsMatcher(r *http.Request, rm *router.RouteMatch) bool {
 }
 
 // NewHTTP returns HTTP server
```
Danielius1922: It doesn't seem like HTTP **server** anymore.

Danielius1922 (approved):

http-gateway/service/httpApi.go
jkralik: fixed

ondrejtomcik (approved):

PR #726, opened 2022-05-11
things_to_devices

PatrikMatiasko:


jkralik:
@Danielius1922 could you look at golangci-lint, why it is a report issue when this PR contains only js?

Danielius1922:
> @Danielius1922 could you look at golangci-lint, why it is a report issue when this PR contains only js?

All of the warnings are in stylecheck linter, there were several linters that didn't work with go1.18, I think they must have fixed it and reenabled it. You can ignore it for now, I'll fix it.

http-gateway/web/public/web_configuration.json
```diff
@@ -1,15 +1,17 @@
 {
-  "domain": "",
-  "httpGatewayAddress": "",
+  "authority": "https://auth.plgd.cloud",
```
kuby: Please make sure to leave this configuration with empty values, this should only be on git as a template. We don't want anyone else to use our clientIds and other params.

http-gateway/web/src/components/menu/menu-item.js
```diff
@@ -12,7 +12,12 @@ export const MenuItem = ({
   to,
   ...rest
 }) => {
-  const menuItemClassName = classNames('menu-item', className)
+  const location = useLocation()
+  const menuItemClassName = classNames(
+    'menu-item',
+    className,
+    to === '/' && location.pathname.includes('devices') ? 'active' : ''
```
kuby: Could you please explain the need for this addition here? Also from line 36.

As far as I remember, the `NavLink` has a prop `activeClassName` which is by default `active`, so this really should be done automatically (even though I know that in react-router v6, this prop will be deleted). I am more specifically curious about this `location.pathname.includes('devices')`, why is this hardcoded into the `MenuItem` component? If you really need this for some reason (meaning the `exact` is not enough anymore), you can pass this from the `<Menu>` component like:

```javascript
<MenuItem
   to="/"
   icon="fa-list"
   tooltip={collapsed && _(t.devices)}
   className={classNames({ 'active': location.pathname.includes('devices') })}
>
   {_(t.devices)}
</MenuItem>
```

http-gateway/web/src/index.js
```diff
@@ -37,7 +37,7 @@ fetch('/web_configuration.json')
       // only set the code to the session storage, so that the caller can process it.
       const urlParams = new URLSearchParams(window.location.search)
       const code = urlParams.get('code')
-      if (window.location.pathname === '/things' && code) {
+      if (window.location.pathname === '/' && code) {
```
kuby: This is the part which I mentioned you in the chat. Please, change the `/` to `/devices`, I've already registered this URL in the client so it will work (at least locally). To give you a bit more about where it comes from, this is the code which renders an iframe with the provided URL (which contains the /devices) https://github.com/plgd-dev/hub/pull/726/files#diff-98731371d18057ad29d7bb7d0f04a52fb2f0da9f2366e8c4ebe7c958d517f136R150 .

Notice that the URL contains a get parameter `code`, which indicated that we are requesting an authorization code. This goes through the `deviceOauthClient`, that's why we have to make sure, this URL is registered in the client as a redirect URI.

This `if` statement here makes sure, that if the URL does contain `/devices` and `code` as a get parameter (which it will, because that's how we get redirected back from the auth server), it stores the `code` value into the sessions storage, and the Promise which rendered the iframe will check for few seconds for that stored variable. This was the only way I was able to implement this, cause the API is not fetchable...

@ondrejtomcik Make sure you update the clients later.

http-gateway/web/src/containers/devices/provision-new-device.js
```diff
@@ -11,7 +11,7 @@ import { CopyBox } from '@/components/copy-box'
 import { isValidGuid } from '@/common/utils'
 
 import { getDeviceAuthCode } from './rest'
-import { messages as t } from './things-i18n'
+import { messages as t } from './devices-i18n'
 
 const NOOP = () => {}
```
kuby, line 17: While I was testing this just now, I found a small mistake in the code, I want to ask you to correct it, it is in this file on lines 143., 150. and the other divs, which has the same `id` `auth-code-box`. I think you can change it to `className`, also in the `device-details.scss` file.

One more addition to that class to add could be this:
```css
.auth-code-box {
   span {
      text-align: right;
   }
}
```

Thanks!

kuby (changes requested):

http-gateway/web/src/components/menu/menu-item.js
PatrikMatiasko: I agree -> It can be set as a prop. 
Why I need this -> It comes from concept: 

Menu with link "to" is not active only for "/" but also for "/devices/X/Y. So active is not only for home route but also for details of devices.

http-gateway/web/public/web_configuration.json
PatrikMatiasko: Is this good adept to gitignore ? Maybe we can transform this file into .env only for current user without reverting before commit ;)

http-gateway/web/src/index.js
ondrejtomcik: Will do

http-gateway/web/src/index.js
```diff
@@ -37,7 +37,7 @@ fetch('/web_configuration.json')
       // only set the code to the session storage, so that the caller can process it.
       const urlParams = new URLSearchParams(window.location.search)
       const code = urlParams.get('code')
-      if (window.location.pathname === '/things' && code) {
+      if (window.location.pathname === '/devices' && code) {
```
ondrejtomcik, line 40: Update also code comment :)

ondrejtomcik:
@PatrikMatiasko I pushed one commit to your branch, to update deployment configuration - https://github.com/plgd-dev/hub/pull/726/commits/341dfaa0b4149f3c14269b8daa5a6c0ca6bc32d1

http-gateway/web/public/web_configuration.json
kuby: Well, we could certainly add this file to .gitignore.

Regarding the .env, the idea of this JSON config being placed into the public folder and fetched on the start of the application was that we are able to change this config after the app is built, since this is a static resource not bundled with the app. With .env, we could not achieve this anymore and the params would be "hardcoded" during the build pahse.

I'd keep this now as it is.

http-gateway/web/src/containers/devices/_command-timeout-control.js
```diff
@@ -47,10 +47,10 @@ export const CommanTimeoutControl = ({
   )
 
   const units = Object.values(commandTimeoutUnits)
-    .filter(unit => unit !== NS)
-    .map(unit => ({
-      value: unit,
-      label: unit === INFINITE ? _(t.default) : unit,
+    .filter(_unit => _unit !== NS)
```
kuby, line 50: Can you explain this change? I really don't see any reason why to change this to underscores :)

PatrikMatiasko: Sonar ->” variable is already declared in the upper scope “

kuby: Ah, I did not see it, sorry :)

kuby (approved):

PR #735, opened 2022-05-24
coap-gateway: check and unpublish resources if the device doesn't contain them

jkralik:


coap-gateway/service/observation/deviceObserver.go
```diff
@@ -327,6 +375,39 @@ func getPublishedResources(ctx context.Context, rdClient GrpcGatewayClient, devi
...
+	if len(publishedResources) == 0 {
+		return nil, nil
+	}
+	if len(deviceResources) == 0 {
+		return publishedResources, nil
+	}
+	validResources := make([]*commands.Resource, 0, len(publishedResources))
+	toUnpublishResourceInstanceIds := make([]int64, 0, len(publishedResources))
+	for _, res := range publishedResources {
+		if _, ok := deviceResources.GetResourceLink(res.GetHref()); ok {
```
Danielius1922, line 411: Nitpick: This is O(number of published_resources x number of deviceResources) ~ O(n^2). With ordered arrays / sets we could get it down to O(n)

coap-gateway/service/observation/observation_test.go
```diff
@@ -103,87 +97,22 @@ func TestIsResourceObservableWithInterface(t *testing.T) {
...
-		},
-		{
-			name: testResourceName(device.ResourceURI) + "with not supported observeInterface",
-			args: args{
-				resourceHref:     device.ResourceURI,
-				resourceType:     device.ResourceType,
-				observeInterface: interfaces.OC_IF_B,
-			},
-			want: false,
+			name:    testResourceName("/oic/res"),
```
Danielius1922: "/oic/res" -> resources.ResourceURI

I would also add at least one case with `wantErr` true (invalid href maybe?) and `want` true (some obsevable resource).

coap-gateway/service/observation/deviceObserver_test.go
```diff
@@ -259,8 +414,8 @@ func runTestDeviceObserverRegister(ctx context.Context, t *testing.T, deviceID s
 	// give time to wait for data
 	ctxWait, waitCancel := context.WithTimeout(context.Background(), time.Second*10)
 	closeWaitChans := func() {
-		close(retrieveChan)
-		close(observeChan)
+		//	close(retrieveChan)
```
Danielius1922: Revert?

coap-gateway/service/observation/deviceObserver.go
```diff
@@ -178,12 +216,27 @@ func emptyDeviceIDError() error {
...
+	}
+	resourceHref := resources.ResourceURI
+	observeInterface := interfaces.OC_IF_B
+	res, ok := links.GetResourceLink(resourceHref)
+	if !ok {
+		return false, fmt.Errorf("resourceLink for href(%v) not found", resourceHref)
+	}
+
+	observable := res.Policy.BitMask.Has(schema.Observable)
+	if observeInterface == "" || !observable {
```
Danielius1922: `observeInterface == ""` is never true now

Danielius1922 (changes requested):

coap-gateway/service/observation/deviceObserver.go
jkralik: fixed

coap-gateway/service/observation/observation_test.go
jkralik: done

coap-gateway/service/observation/deviceObserver_test.go
jkralik: fixed

coap-gateway/service/observation/deviceObserver.go
jkralik: fixed

coap-gateway/service/observation/deviceObserver.go
```diff
@@ -375,15 +375,35 @@ func getPublishedResources(ctx context.Context, rdClient GrpcGatewayClient, devi
...
+	validResources = make([]*commands.Resource, 0, len(publishedResources))
+	toUnpublishResourceInstanceIds = make([]int64, 0, len(publishedResources))
+	publishedResources.Sort()
+	deviceResources.Sort()
+	var j int
+	for _, res := range publishedResources {
+		for j < len(deviceResources) && deviceResources[j].Href < res.GetHref() {
+			j++
+		}
+		if deviceResources[j].Href == res.GetHref() {
```
Danielius1922, line 391: In the case that `publishedResources` has a resource with >Href than the last element of `deviceResources` then `j=len(deviceResources)`, which means `deviceResources[j].Href` will be out-of-bounds. 

The simple fix is to add
```
if j>=len(deviceResources) {
  break
}
```
after the inner loop.

coap-gateway/service/observation/observation.go
```diff
@@ -6,15 +6,14 @@ import (
 
 	"github.com/plgd-dev/device/schema"
 	"github.com/plgd-dev/device/schema/interfaces"
-	"github.com/plgd-dev/device/schema/resources"
 	coapMessage "github.com/plgd-dev/go-coap/v2/message"
 	"github.com/plgd-dev/go-coap/v2/message/codes"
 	"github.com/plgd-dev/kit/v2/codec/cbor"
 )
 
 // Query /oic/res resource to get device resource links.
```
Danielius1922: Docu no longer correct, other resources that support `if.ll` interface can now be queried too.

coap-gateway/service/observation/deviceObserver.go
jkralik: fixed

coap-gateway/service/observation/observation.go
jkralik: fixed

Danielius1922 (approved):

PR #736, opened 2022-05-26
resource-directory: dont use RWmutex as recursive mutex - causes deadlock

jkralik:
…lock

### DIFF
--- a/resource-aggregate/cqrs/eventstore/projection.go
+++ b/resource-aggregate/cqrs/eventstore/projection.go
@@ -174,11 +174,13 @@ func (i *iterator) Err() error {
 	return i.iter.Err()
 }
 
-func (p *Projection) getModelLocked(ctx context.Context, groupID, aggregateID string) (*aggregateModel, error) {
+func (p *Projection) getModel(ctx context.Context, groupID, aggregateID string) (*aggregateModel, error) {
 	var ok bool
 	var mapApm map[string]*aggregateModel
 	var apm *aggregateModel
 
+	p.lock.Lock()
+	defer p.lock.Unlock()
 	if mapApm, ok = p.aggregateModels[groupID]; !ok {
 		mapApm = make(map[string]*aggregateModel)
 		p.aggregateModels[groupID] = mapApm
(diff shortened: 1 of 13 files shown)


resource-aggregate/cqrs/eventstore/projection.go
```diff
@@ -289,44 +289,70 @@ func (p *Projection) Forget(queries []SnapshotQuery) (err error) {
...
 
-func (p *Projection) allModelsLocked(onModel func(m Model) (wantNext bool)) {
+func (p *Projection) iterateOverAllModels(onModel func(m Model) (wantNext bool)) {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
 	for _, group := range p.aggregateModels {
 		for _, apm := range group {
-			if !onModel(apm.model) {
-				return // stop iteration
+			p.lock.RUnlock()
```
Danielius1922, line 297: This works, but instead of locking and unlocking in each step wouldn't something like this be better:

```
p.lock.RLock()
create deep copy of p.aggregateModels
p.lock.RUnlock()

iterate over the copy
```

It would reduce the contention on the lock I think.

jkralik: This is what I want to avoid.  `aggregateModels` can contains 500K elements so copy is to expensive on CPU.

jkralik: mainly allocation

resource-directory/service/deviceMetadataProjection.go
```diff
@@ -10,24 +12,45 @@ import (
 
 // protected by lock in Projection struct in resource-aggregate/cqrs/eventstore/projection.go
 type deviceMetadataProjection struct {
-	data *events.DeviceMetadataSnapshotTaken
+	deviceID string
+
+	private struct {
```
Danielius1922, line 17: Is this a Go v1.18 new feature? I wasn't aware that `private` keyword even existed in golang!

Danielius1922, line 17: I like it, this way the data and the lock are obviously paired!

Danielius1922 (approved):

PR #744, opened 2022-06-08
grpc-gateway: extend Device proto message

jkralik:
need for client-application


resource-directory/service/deviceDirectory.go
```diff
@@ -69,9 +70,11 @@ func decodeContent(content *commands.Content, v interface{}) error {
 }
 
 type Device struct {
-	ID       string
-	Resource *device.Device
-	Metadata *pb.Device_Metadata
+	ID              string
+	Resource        *device.Device
```
Danielius1922: Bit weird that variable named `Resource` is of type `Device`. :D

Danielius1922 (approved):

resource-directory/service/deviceDirectory.go
ondrejtomcik: Hm yes. Jozef? 😂

ondrejtomcik (approved):

PR #745, opened 2022-06-14
services: share file watcher among cert managers

jkralik:


pkg/security/certManager/general/certManager.go
```diff
@@ -40,56 +41,55 @@ func (c Config) Validate() error {
...
-	doneWg                  sync.WaitGroup
-	done                    chan struct{}
+	fileWatcher             *fsnotify.Watcher
 	verifyClientCertificate tls.ClientAuthType
 	logger                  log.Logger
 
-	mutex      sync.Mutex
-	tlsKeyPair tls.Certificate
-	tlsCAPool  *x509.CertPool
+	mutex            sync.Mutex
```
Danielius1922: inner struct for mutex and members protected by the mutex? it's no longer clear which those are now

jkralik: fixed

Danielius1922 (approved):

PR #763, opened 2022-08-03
Enable linters in golansci-lint and fix issues

Danielius1922:
Enabled linters:
    - nakedret
    - nilerr

### DIFF
--- a/http-gateway/serverMux/jsonMarshaler.go
+++ b/http-gateway/serverMux/jsonMarshaler.go
@@ -7,6 +7,7 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/plgd-dev/go-coap/v2/message"
+	"github.com/plgd-dev/hub/v2/pkg/log"
 	"github.com/plgd-dev/kit/v2/codec/cbor"
 	"github.com/plgd-dev/kit/v2/codec/json"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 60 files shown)


jkralik (approved):

http-gateway/serverMux/jsonMarshaler.go
```diff
@@ -135,13 +135,13 @@ func modify(v interface{}) (newValue interface{}, wantDelete bool) {
...
 	if err != nil {
-		return data, err
+		return nil, err
 	}
 
 	var val interface{}
 	err = json.Decode(data, &val)
 	if err != nil {
-		return data, nil
+		return nil, err
```
jkralik: `return data, nil`

pkg/log/log.go
```diff
@@ -369,7 +369,7 @@ func (l *WrapSuggarLogger) Fatalf(template string, args ...interface{}) {
 
 func (l *WrapSuggarLogger) LogAndReturnError(err error) error {
 	if err == nil {
-		return err
+		return nil
```
jkralik, line 372: return `data, nil`

jkralik (changes requested):
look to comments

http-gateway/serverMux/jsonMarshaler.go
Danielius1922: Only this line? What about line 157 I think we talked yesterday about that as well.

Danielius1922: Also, I'd like to add comment to the code. But ...I forgot it. :D Could you please formulate a short comment about why we return data in this case?

jkralik (approved):

PR #764, opened 2022-08-05
coap-gateway: DTLS protocol

jkralik:


coap-gateway/coapconv/grpc.go
```diff
@@ -0,0 +1,58 @@
...
+		return codes.InvalidArgument
+	case coapCodes.InternalServerError:
+		return codes.Internal
+	case coapCodes.NotImplemented:
+		return codes.Unimplemented
+	case coapCodes.BadGateway:
+	case coapCodes.ServiceUnavailable:
+		return codes.Unavailable
+	case coapCodes.GatewayTimeout:
+	case coapCodes.ProxyingNotSupported:
```
Danielius1922: Is there a special value for these two last items? If not then they can be removed from the switch and line 57 will get called anyway.

Danielius1922 (approved):

coap-gateway/coapconv/grpc.go
jkralik: Fixed

Danielius1922 (approved):

PR #769, opened 2022-08-16
coap-gw: set not found for empty body from batch observation

### LINKED ISSUE #766: Recreated resource is not republished
A resource was created with specific attributes and values. Then it was deleted and re-created again with exact same properties as before.
It was observed that resource was not republished.

If the resource is re-created but with a different value in some attribute, it is republished.

jkralik:
ra: set not found to unpublished resources

Fixes #766

### DIFF
--- a/grpc-gateway/service/createAndDeleteResource_test.go
+++ b/grpc-gateway/service/createAndDeleteResource_test.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"os"
 	"testing"
+	"time"
 
 	"github.com/plgd-dev/device/schema"
 	"github.com/plgd-dev/device/schema/interfaces"
@@ -160,7 +160,7 @@ func createSwitchResourceExpectedEvents(t *testing.T, deviceID, subID, correlati
 	}
 }
 
-func deleteSwitchResourceExpectedEvents(t *testing.T, deviceID, subID, correlationID, switchID string, isDiscoveryResourceBatchObservable bool) map[string]*pb.Event {
+func deleteSwitchResourceExpectedEvents(t *testing.T, deviceID, subID, correlationID, switchID string) map[string]*pb.Event {
 	deletePending := &pb.Event{
 		SubscriptionId: subID,
 		CorrelationId:  correlationID,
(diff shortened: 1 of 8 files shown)


ondrejtomcik (approved):

grpc-gateway/service/createAndDeleteResource_test.go
```diff
@@ -281,15 +270,14 @@ func TestCreateAndDeleteResource(t *testing.T) {
 	subClient, subID := subscribeToAllEvents(t, ctx, c, correlationID)
 	const switchID = "1"
 
-	isDiscoveryResourceBatchObservable := test.IsDiscoveryResourceBatchObservable(ctx, t, deviceID)
-
 	for i := 0; i < 5; i++ {
 		fmt.Printf("iteration %v\n", i)
+		time.Sleep(time.Millisecond * 500)
```
Danielius1922, line 275: Sleep after print seems weird, is the placement correct? Usually we delay after a create/delete call.

jkralik: As it is async, unpublish events can still be processed during resource creation. (some time the iotivity-lite test fails so I needed add there some time to get idle state)

resource-aggregate/events/resourceStateSnapshotTaken.go
```diff
@@ -280,7 +280,6 @@ func (e *ResourceStateSnapshotTaken) HandleEventResourceChanged(ctx context.Cont
 
 func (e *ResourceStateSnapshotTaken) HandleEventResourceDeleted(ctx context.Context, deleted *ResourceDeleted) error {
 	if deleted.GetStatus() == commands.Status_OK || deleted.GetStatus() == commands.Status_ACCEPTED {
-		e.LatestResourceChange = nil
```
Danielius1922, line 283: I recall adding this line a few months ago, won't it be a problem if we keep the resource change after the resource is deleted?

jkralik: When resource has been deleted we expect two same events that set `LatestResourceChange`:

1. unpublish event generates `ResourceContentChanged` event with status `NotFound`
2. cancellation of observation  generates `ResourceContentChanged` event with status `NotFound`

It depends which comes first do this is why it need to be same for deduplication.

Danielius1922 (approved):

PR #773, opened 2022-09-07
http-gateway: propagate logger to logging middleware

jkralik:
- add error key to http log

### DIFF
--- a/pkg/net/http/loggingMiddleware.go
+++ b/pkg/net/http/loggingMiddleware.go
@@ -8,15 +8,21 @@ import (
 	"strings"
 	"time"
 
+	"github.com/plgd-dev/go-coap/v2/message"
+	"github.com/plgd-dev/hub/v2/http-gateway/uri"
 	"github.com/plgd-dev/hub/v2/pkg/log"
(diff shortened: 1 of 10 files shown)


pkg/net/http/loggingMiddleware.go
```diff
@@ -28,7 +34,25 @@ func (w *statusWriter) Write(b []byte) (int, error) {
...
+		switch w.ResponseWriter.Header().Get(uri.ContentTypeHeaderKey) {
+		case uri.ApplicationProtoJsonContentType:
+		case message.AppJSON.String():
+			var s rpcStatus.Status
+			if err := protojson.Unmarshal(b, &s); err == nil {
+				w.err = status.ErrorProto(&s)
+			}
+		}
+		if w.err == nil {
+			errLen := 1024
```
Danielius1922, line 47: Is the header limited to this size or why the truncation of the error?

Danielius1922 (approved):

pkg/net/http/loggingMiddleware.go
jkralik: I want to just limit size of error message - it can be big

PR #774, opened 2022-09-09
certificate-authority: integrate http service for web apps

jkralik:


certificate-authority/config.yaml
```diff
@@ -41,6 +41,12 @@ apis:
           keyFile: "/secrets/private/cert.key"
           certFile: "/secrets/public/cert.crt"
           useSystemCAPool: false
+  http:
+    address: "0.0.0.0:9101"
+    readTimeout: 8s
+    readHeaderTimeout: 4s
+    writeTimeout: 16s
+    idleTimeout: 30s
```
ondrejtomcik, line 49: tls and authorization are from gRPC are applied to http?

ondrejtomcik (approved):

certificate-authority/config.yaml
jkralik: Yes

grpc-gateway/service/service.go
```diff
@@ -47,13 +47,13 @@ func New(ctx context.Context, config Config, fileWatcher *fsnotify.Watcher, logg
 
 	pool, err := ants.NewPool(config.Clients.Eventbus.GoPoolSize)
 	if err != nil {
-		server.Close()
+		_ = server.Close()
```
Danielius1922: This could be problematic if the implementation ever changes and returns a real error. I'm 99% sure that the implementer won't think about needing to update all calls of `server.Close` to check error.

Danielius1922 (approved):

grpc-gateway/service/service.go
jkralik: fixed

Danielius1922: There are `s.Close()` and `grpcServer.Close()` calls outside testing code that ignore the error. :innocent:

jkralik: ok fixed :)

Danielius1922: Look good. ;)

Danielius1922 (approved):

PR #788, opened 2022-09-14
http-gateway: introduce /.well-known/configuration endpoint

jkralik:


bundle/nginx/nginx.conf.template
```diff
@@ -93,7 +93,21 @@ http {
...
+      proxy_ssl_trusted_certificate   /data/certs/root_ca.crt;
+      proxy_ssl_verify                on;
+      set $cors_headers 'Authority,Method,Path,Scheme,Accept,Accept-Encoding,Accept-Language,Content-Type,Origin,Refer,Sec-Fetch-Dest,Sec-Fetch-Mode,Sec-Fetch-Site,Authorization,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,correlation-id';
+      include /nginx/cors.conf;
+      proxy_http_version 1.1;
+      proxy_set_header Upgrade $http_upgrade;
+      proxy_set_header Connection $connection_upgrade;
+      proxy_set_header Host $host;
+    }
+    location ~ ^(/|/api/v1/devices|/api/v1/ws/devices|/.well-known/hub-configuration|/.well-known/configuration|/auth_config.json) {
```
ondrejtomcik: what is actually auth_config.json?

jkralik: removed

ondrejtomcik (approved):

resource-directory/test/service.go
```diff
@@ -30,7 +30,7 @@ func MakeConfig(t *testing.T) service.Config {
 	cfg.Clients.Eventstore.ProjectionCacheExpiration = time.Second * 120
 
 	cfg.ExposedHubConfiguration.CAPool = config.CA_POOL
-	cfg.ExposedHubConfiguration.AuthorizationServer = "https://" + config.OAUTH_SERVER_HOST
+	cfg.ExposedHubConfiguration.Authority = "https://" + config.OAUTH_SERVER_HOST
```
ondrejtomcik, line 33: Perfect that it runs in tests but it's not working in real. Bundle, hub nor helm charts. We should reconsider how we run the tests, as they were useless and the current main is not usable and we didn't discover it before merge.

PR #793, opened 2022-09-24
Various UI fix

PatrikMatiasko:
Solves:

- Resource Interface Dropdown not working and has wrong naming
- Profile picture size doesn't fit 
- Retrieve on resource in UI returns wrongly formatted model


ondrejtomcik (approved):

http-gateway/web/src/components/table/utils.js
```diff
@@ -1,3 +1,7 @@
 // Case insensitive sort function
-export const compareIgnoreCase = (a, b) =>
-  a.localeCompare(b, 'en', { numeric: true, sensitivity: 'base' })
+export const compareIgnoreCase = (a, b) => {
+  const item1 = Array.isArray(a) ? a[0] : a
+  const item2 = Array.isArray(b) ? b[0] : b
+
+  return item1.localeCompare(item2, 'en', { numeric: true, sensitivity: 'base' })
```
jkralik: for me it is better when you `sort` array and then one by one compare values.
eg for input:
```
["a", "b"]
["a"]
```
I expect the result:
```
["a"]
["a", "b"]
```

jkralik (changes requested):
Fix sorting.

http-gateway/web/src/components/table/utils.js
PatrikMatiasko: I just fix issue with sorting, but we can do it definitely better if you want.

jkralik (approved):

PR #795, opened 2022-09-30
coap-gateway: allow DTLS protocol

jkralik:


test/config/config.go
```diff
@@ -46,15 +46,24 @@ const (
...
-	NATS_URL    = "nats://localhost:4222"
-	OWNER_CLAIM = "sub"
+	CA_POOL                  = os.Getenv("LISTEN_FILE_CA_POOL")
+	KEY_FILE                 = os.Getenv("LISTEN_FILE_CERT_DIR_PATH") + "/" + os.Getenv("LISTEN_FILE_CERT_KEY_NAME")
+	CERT_FILE                = os.Getenv("LISTEN_FILE_CERT_DIR_PATH") + "/" + os.Getenv("LISTEN_FILE_CERT_NAME")
+	MONGODB_URI              = "mongodb://localhost:27017"
+	NATS_URL                 = "nats://localhost:4222"
+	OWNER_CLAIM              = "sub"
+	COAP_GATEWAY_UDP_ENABLED = os.Getenv("TEST_COAP_GATEWAY_UDP_ENABLED") == "true"
+	USE_COAP_SCHEME          = func() string {
```
Danielius1922: USE_COAP_SCHEME -> použi coap schému, to je naschvál? Čakal by som skôr USED_COAP_SCHEME (=použitá coap schéma) alebo možno ACTIVE_COAP_SCHEME (=aktívna coap schéma).

test/iotivity-lite/service/republish_test.go
```diff
@@ -141,7 +141,7 @@ func TestRepublishAfterRefresh(t *testing.T) {
 	}()
 	c := pb.NewGrpcGatewayClient(conn)
 
-	_, shutdownDevSim := test.OnboardDevSim(ctx, t, c, deviceID, config.GW_HOST, nil)
+	_, shutdownDevSim := test.OnboardDevSim(ctx, t, c, deviceID, config.COAPS_TCP_SCHEME+config.COAP_GW_HOST, nil)
```
Danielius1922: Why do some calls use `config.COAPS_TCP_SCHEME` (like here), but others (like in test/pb/pendingCommand.go) use `config.USE_COAP_SCHEME`? Is there a method to it? Can't  `config.USE_COAP_SCHEME` be used everywhere? It should have the correct value based on env settings.

go.mod
```diff
@@ -21,9 +21,11 @@ require (
 	github.com/lestrrat-go/jwx v1.2.25
 	github.com/nats-io/nats.go v1.16.0
 	github.com/panjf2000/ants/v2 v2.5.0
+	github.com/pion/dtls/v2 v2.1.6-0.20220815090156-a04cfccb23e5
+	github.com/pion/logging v0.2.2
 	github.com/plgd-dev/device/v2 v2.0.0-20220929115241-033fe063adb5
 	github.com/plgd-dev/go-coap/v2 v2.6.1-0.20220801124204-0ab75f6e0680
```
Danielius1922: Shouldn't this branch only have `go-coap/v3` instaed of both v2 and v3?

pkg/net/coap/service/config.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if c.KeepAlive == nil && c.InactivityMonitor == nil {
+		return fmt.Errorf("keepAlive or inactivityMonitor must be set")
+	}
+
+	return nil
+}
+
+type TLSConfig struct {
+	Enabled                        *bool                    `yaml:"enabled,omitempty" json:"enabled,omitempty"`
```
Danielius1922, line 82: We need 3 values or why else the pointer instead of value here?

jkralik: I don't show these values when print configuration (hidden) - `omitempty` works only when it is enabled.
For eg: For DPS `Enabled` doesn't have sense in configuration. So in config.yaml and I don't show this attribute.
Similar for `MonitoringInactivity` (DPS) and `KeepAlive` (CoapGW)

go.mod
jkralik: Fixed

test/iotivity-lite/service/republish_test.go
jkralik: Fixed

test/config/config.go
jkralik: Fixed

Danielius1922 (approved):

charts/plgd-hub/values.yaml
```diff
@@ -712,20 +712,36 @@ coapgateway:
   # -- Annotations for coap-gateway pod
   podAnnotations: { }
   service:
-    # -- Service type
-    type: LoadBalancer
```
ondrejtomcik, line 716: This is a breaking change. CoAP Gateway by default doesn't go through the LoadBalancer, so users need to configure the deployment as documented here:
https://docs.plgd.dev/docs/deployment/k8s/#nodeport-for-coap-gateway
This won't work anymore and update of deployment will stop working. 

You don't need to configure the `nodeport` per protocol. Nor the `type`. I would keep the old configuration for the service, (don't change `values.yaml` at all) and create `service-tcp.yaml` if tcp is in protocols, create `service-udp.yaml` if it is listed in protocols, but reuse same `labels`, `annotation`, `node port` as well as `type`.

jkralik: Ok so I suggest to consider `service` at TCP setup and creates `serviceUdp` - this will be backward compatible.

ondrejtomcik (approved):

PR #814, opened 2022-11-09
certManager: support string array and string in caPool property

jkralik:


ondrejtomcik (approved):
Please update documentation of config properties in doc. it's important for auth.caPool

pkg/security/certManager/client/certManager_test.go
```diff
@@ -45,6 +46,93 @@ AnQ1eTGKTGLdAZsV+NnZPL17nit1cbiN2g==
...
+	}{
+		{
+			name: "caPool as string",
+			args: args{
+				configYaml: `caPool: /tmp/test3570354545/ca3202122454
+keyFile: /tmp/test3570354545/key3658110735
+certFile: /tmp/test3570354545/crt4065348335
+`,
+			},
+			want: client.Config{
```
Danielius1922, line 68: tt.want is unused as far as I can see

pkg/security/certManager/client/certManager.go
```diff
@@ -7,18 +7,26 @@ import (
...
 type Config struct {
-	CAPool          string `yaml:"caPool" json:"caPool" description:"file path to the root certificates in PEM format"`
-	KeyFile         string `yaml:"keyFile" json:"keyFile" description:"file name of private key in PEM format"`
-	CertFile        string `yaml:"certFile" json:"certFile" description:"file name of certificate in PEM format"`
-	UseSystemCAPool bool   `yaml:"useSystemCAPool" json:"useSystemCaPool" description:"use system certification pool"`
+	CAPool          interface{} `yaml:"caPool" json:"caPool" description:"file path to the root certificates in PEM format"`
+	KeyFile         string      `yaml:"keyFile" json:"keyFile" description:"file name of private key in PEM format"`
+	CertFile        string      `yaml:"certFile" json:"certFile" description:"file name of certificate in PEM format"`
+	UseSystemCAPool bool        `yaml:"useSystemCAPool" json:"useSystemCaPool" description:"use system certification pool"`
+	CAPoolArray     []string    `yaml:"-" json:"-"`
```
Danielius1922: I don't like that this is exported, this is an internal field generated from the `CAPool` field.

What about something like this:

```go
type Config struct {
...
  caPoolArray []string `yaml:"-" json:"-"`...
  validated bool
}

func (c *Config) CAPoolArray()([]string, error) {
  if !c.validated {
    return nil, error
  }
  return c.caPoolArray, nil
}
```

pkg/security/certManager/client/certManager_test.go
jkralik: fixed :)

pkg/security/certManager/general/certManager_test.go
```diff
@@ -113,15 +113,15 @@ func createTmpCertFiles(t *testing.T, caFile, crtFile, keyFile string) general.C
...
+		CAPool:   []string{caFile},
 		KeyFile:  keyFile,
 		CertFile: crtFile,
 	}
 	return cfg
 }
 
 func deleteTmpCertFiles(t *testing.T, cfg general.Config) {
-	err := os.Remove(cfg.CAPool)
+	err := os.Remove(cfg.CAPool[0])
```
Danielius1922: I would range over the array here, that way the code will continue to work if we ever modify the constructor to create CAPool with multiple files.

jkralik: fixed

pkg/security/certManager/client/certManager.go
jkralik: fixed

Danielius1922: yep, good job.

Danielius1922 (approved):

PR #816, opened 2022-11-10
resource-aggregate: fix race in DeviceMetadataSnapshotTaken

jkralik:
events must be immutable

### DIFF
--- a/resource-aggregate/events/deviceMetadataSnapshotTaken.go
+++ b/resource-aggregate/events/deviceMetadataSnapshotTaken.go
@@ -84,11 +84,7 @@ func (d *DeviceMetadataSnapshotTaken) HandleDeviceMetadataUpdated(ctx context.Co
 	}
 
 	d.DeviceId = upd.GetDeviceId()
-	if d.DeviceMetadataUpdated == nil {
-		d.DeviceMetadataUpdated = upd
-	} else {
-		d.DeviceMetadataUpdated.CopyData(upd)
-	}
+	d.DeviceMetadataUpdated = upd
 	d.EventMetadata = upd.GetEventMetadata()
 	return true, nil
 }


resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -84,11 +84,7 @@ func (d *DeviceMetadataSnapshotTaken) HandleDeviceMetadataUpdated(ctx context.Co
 	}
 
 	d.DeviceId = upd.GetDeviceId()
-	if d.DeviceMetadataUpdated == nil {
-		d.DeviceMetadataUpdated = upd
-	} else {
-		d.DeviceMetadataUpdated.CopyData(upd)
-	}
+	d.DeviceMetadataUpdated = upd
```
Danielius1922, line 87: We still have a bit of dangerous code relating to the `DeviceMetadataUpdated` struct:

```
func (p *deviceMetadataProjection) GetDeviceMetadataUpdated() *events.DeviceMetadataUpdated {
	p.private.lock.RLock()
	defer p.private.lock.RUnlock()
	return p.private.snapshot.GetDeviceMetadataUpdated()
}
```

This only locks the snapshot and it is still possible to get the same `DeviceMetadataUpdated` pointer in multiple threads. Only if the pointers are used as read-only (eg the member variables are never modified) it is safe, but there is no `const` in golang, so we cannot guarantee it. To make it 100% safe we would have to return a deep copy from `GetDeviceMetadataUpdated() `.

Danielius1922 (approved):

resource-aggregate/events/deviceMetadataSnapshotTaken.go
jkralik: `p.private.snapshot.DeviceMetadataUpdated` must be immutable and the `GetDeviceMetadataUpdated()` is generated by protoc.

PR #834, opened 2022-12-08
grpc-gateway: extend hub configuration

jkralik:


grpc-gateway/pb/hubConfiguration_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeviceOAuthClient_Clone(t *testing.T) {
+	a := DeviceOAuthClient{
+		ProviderName: "test",
+		Scopes:       []string{"test"},
+	}
+	b := a.Clone()
+	require.Equal(t, &a, b)
```
Danielius1922, line 16: The intent of the Clone functions seems to be to do a deep copy, right? I'd extend the tests to check that the variables do not share pointers, thus modifying `a` has no effect on `b`.

jkralik: strings can share the pointers

jkralik: only structs needs to copied

jkralik: added check for arrays

Danielius1922: > strings can share the pointers

Yes, but if you change the string in a it doesn't effect the string in b. Apparently, strings and arrays will have a Clone method in golang v1.20.
Though string was never the problem. The problem I had it mind was when you modify one array, it should affect the other array as well, if recall golang behavior correctly. So something like this would be a problem with just an assignment.

```go
a := DeviceOAuthClient{
	ProviderName: "test",
	Scopes:       []string{"a", "a", "a"},
}
b := a.Clone()
b.Scopes[0] = "b"
require.NotEqual(t, a, b)
```

Danielius1922 (approved):

grpc-gateway/pb/hubConfiguration_test.go
jkralik: 👍

resource-aggregate/events/deviceMetadataSnapshotTaken.go
```diff
@@ -300,6 +300,9 @@ func (d *DeviceMetadataSnapshotTaken) updateDeviceConnection(ctx context.Context
 		return nil, status.Errorf(codes.InvalidArgument, "cannot update connection status for empty connectionId")
 	}
 	if !req.GetConnection().IsOnline() {
+		if em.GetVersion() == 0 {
```
Danielius1922, line 303: Version value starts from 1 if set? 

Otherwise we cannot distinguish been when `em` is `nil` and `em` is not `nil`, but `version` is equal to 0.

PR #849, opened 2023-01-13
grpc/nats: log body of requests/responds/messages

jkralik:
- added `log.dumpBody` to all services
- grpc unary/stream calls logs request/responds/stream messages
- nats publish logs messages


pkg/net/grpc/server/logInterceptors.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	return logger
+}
+
+func (w *logServerStream) RecvMsg(m interface{}) error {
+	err := w.ServerStream.RecvMsg(m)
+	code := status.Code(err)
+	level := DefaultCodeToLevel(code)
+	if errors.Is(err, io.EOF) {
+		return err
```
Danielius1922, line 256: The code in `SendMsg` is very similar, expect there the code is like this:

```go
	if errors.Is(err, io.EOF) {
		level = log.DebugLevel
	}
````

Is the difference deliberate?

jkralik: Yes. When recv gets error EOF it means that client sends close message. In send if you get EOF this means that you was not able to send message to client so we want to log which message was not send.

Danielius1922 (approved):

PR #911, opened 2023-03-17
coap-gateway: allow to set identityPropertiesRequired for TLS connection 

jkralik:
When identityPropertiesRequired: false, then the device need to use the
certificate with requirements:
 - client, server EKUs
 - must be valid at the current time


coap-gateway/service/config.go
```diff
@@ -93,13 +95,65 @@ func (c *AuthorizationConfig) Validate() error {
...
+
+func (c *COAPConfigMarshalerUnmarshaler) UnmarshalYAML(value *yaml.Node) error {
+	err := value.Decode(&c.COAPConfig)
+	if err != nil {
+		return err
+	}
+	return value.Decode(&c.InjectedCOAPConfig)
+}
+
+func (c COAPConfigMarshalerUnmarshaler) MarshalYAML() (interface{}, error) {
```
Danielius1922, line 132: What's the difference between adding `nonIdentityCertificateEnabled` this way vs. simply adding it to COAPConfig?

jkralik: We don't want to change generic `COAPConfig` so we need to inject property to yaml file to proper place.

Danielius1922 (approved):

PR #920, opened 2023-04-12
Fix security issues reported by SonarCloud

Danielius1922:
Avoid installing package dependencies that are not strictly required by apt and npm in Dockerfiles.

### DIFF
--- a/http-gateway/Dockerfile
+++ b/http-gateway/Dockerfile
@@ -17,7 +17,7 @@ RUN cd /web && \
     npm config set fetch-retry-mintimeout 600000 &&  \
     npm config set fetch-retry-maxtimeout 1200000 && \
     npm config set fetch-timeout 1800000 && \
-    npm install && \
+    npm install --ignore-scripts && \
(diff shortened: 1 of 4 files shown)


Danielius1922:
https://sonarcloud.io/project/security_hotspots?id=plgd-dev_cloud

http-gateway/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN cd /web && \
     npm config set fetch-retry-mintimeout 600000 &&  \
     npm config set fetch-retry-maxtimeout 1200000 && \
     npm config set fetch-timeout 1800000 && \
-    npm install && \
```
Danielius1922, line 20: @jkralik I've tested bundle with this change, seemed OK, is there a quick way to test if this doesn't break http-gw?

jkralik: This is fine - it is used only for build react app

jkralik (approved):

PR #921, opened 2023-04-18
Add option to use custom validFrom timestamp by the test sdkClient

Danielius1922:


test/sdkclient.go
```diff
@@ -80,6 +81,14 @@ func WithRootCA(certificate, key []byte) Option {
 	})
 }
 
+// WithID creates Option that overrides the validFrom timestamp used by the SDK client
+// when generating a certificate use by the client.
+func WithValidFrom(validFrom string) Option {
```
jkralik: I want to replace in future `ValidFrom` and `ValidFor` by `notBefore` and `notAfter` as these fields are defined in certificate specification.

Danielius1922: Added a TODO

jkralik (approved):

PR #934, opened 2023-05-04
test: add plgd time resource

jkralik:


Danielius1922 (approved):

http-gateway/service/getDeviceResources_test.go
```diff
@@ -121,6 +122,15 @@ func TestRequestHandlerGetDeviceResources(t *testing.T) {
 						},
 					),
 				},
+				{
+					Types: []string{plgdtime.ResourceType},
```
Danielius1922, line 126: Maybe we can have a single function in the test package that generates these Change events, otherwise finding all places where they are used is difficult.

jkralik: We need to create task for that because we need to stabilize build in iotivity-lite stack. (All PRs are red now for plgd-dev/device).

Danielius1922: Yep, we have desynced repositories - IoTivity + hub without plgdTime, device with it. We can do it in two steps, merge this first to stabilize all repos and then the refactoring comit.

PR #943, opened 2023-05-26
grpc-gateway: create generate doc for GRPC API

jkralik:


Danielius1922:
I think the generated doc.html should be excluded from SC analysis in the sonar-project.properties configuration file.

Danielius1922 (approved):

PR #956, opened 2023-06-08
services: reload certificates when it is changed

jkralik:
Hub stopped working because of expired mTLS certificates, even though the certs were correctly auto-rotated in the k8s cluster.

### DIFF
--- a/charts/plgd-hub/values.yaml
+++ b/charts/plgd-hub/values.yaml
@@ -53,7 +53,6 @@ extraDeploy:
 # -- External nats dependency setup
 nats:
   enabled: true
-  image: ghcr.io/plgd-dev/hub/nats-server-config-reloader:vnext-pr956
   nats:
     tls:
       secret:
@@ -65,13 +64,13 @@ nats:
   cluster:
     enabled: false
     noAdvertise: false
-
   leafnodes:
     enabled: false
     noAdvertise: false
-
   natsbox:
     enabled: false
+  reloader:
+    image: ghcr.io/plgd-dev/hub/nats-server-config-reloader:vnext-pr956
 
 
 # -- External mongodb-replica dependency setup
--- a/charts/plgd-hub/values.yaml
+++ b/charts/plgd-hub/values.yaml
@@ -100,6 +100,8 @@ mongodb:
   extraVolumeMounts:
     - name: mongodb-crt
       mountPath: "/certs"
+    - name: mongodb-cm-crt
+      mountPath: "/certs-original"
   extraVolumes:
     - name: mongodb-crt
       emptyDir: {}
@@ -107,23 +109,6 @@ mongodb:
       secret:
         secretName: mongodb-cm-crt
   initContainers:
-    - name: convert-cm-crt
-      image: docker.io/bitnami/nginx:1.20.2-debian-10-r63
-      imagePullPolicy: "IfNotPresent"
-      volumeMounts:
-        - name: mongodb-crt
-          mountPath: "/certs"
-        - name: mongodb-cm-crt
-          mountPath: "/tmp/certs"
-      command:
-        - sh
-        - "-c"
-        - |
-          /bin/bash <<'EOF'
-          cat /tmp/certs/tls.crt >> /certs/cert.pem
-          cat /tmp/certs/tls.key >> /certs/cert.pem
-          cp /tmp/certs/ca.crt  /certs/ca.pem
-          EOF
     # workaround for https://github.com/bitnami/charts/issues/10264, https://jira.mongodb.org/browse/MONGOSH-1240
     - name: mongo-binary
       image: docker.io/bitnami/mongodb:5.0.10-debian-11-r3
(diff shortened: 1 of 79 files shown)


pkg/fsnotify/fsnotify.go
```diff
@@ -123,23 +161,84 @@ func (w *Watcher) Close() error {
...
 	defer w.finished.Done()
+	ticks := time.NewTicker(time.Second)
+	ticksRunning := true
+	deletedPaths := make([]string, 0, 16)
+
 	for {
 		select {
 		case <-w.done:
 			return
 		case event := <-w.private.w.Events:
```
Danielius1922: Both `w.private.w.Events` and `w.private.w.Errors` are not locked in this loop. Although they both look like channels to me and  probably they aren't reassigned after construction. Usually a `private` sub structure should be locked, so it caught my eye, maybe we should add a comment here.

jkralik: Ok - removed from private :)

pkg/security/certManager/general/certManager.go
```diff
@@ -68,27 +74,27 @@ func New(config Config, fileWatcher *fsnotify.Watcher, logger log.Logger) (*Cert
...
-			return nil, fmt.Errorf("cannot watch CAPool directory(%v): %w", filepath.Dir(ca), err)
+		if err := c.fileWatcher.Add(ca); err != nil {
+			return nil, fmt.Errorf("cannot watch CAPool(%v): %w", ca, err)
 		}
 	}
 	if config.CertFile != "" {
-		if err := c.fileWatcher.Add(filepath.Dir(config.CertFile)); err != nil {
-			return nil, fmt.Errorf("cannot watch CertFile directory(%v): %w", filepath.Dir(config.CertFile), err)
+		if err := c.fileWatcher.Add(config.CertFile); err != nil {
+			return nil, fmt.Errorf("cannot watch CertFile(%v): %w", config.CertFile, err)
```
Danielius1922, line 100: Shouldn't we remove the previously registered observed paths before returning with error?

pkg/security/certManager/general/certManager.go
```diff
@@ -130,124 +143,191 @@ func (a *CertManager) Close() {
...
+		_, err := a.loadCertsLocked()
+		if err != nil {
+			return nil, err
+		}
+	}
+	if a.private.tlsKeyPair == nil {
+		return nil, fmt.Errorf("certificate is not loaded")
+	}
+
+	return a.private.tlsKeyPair, nil
```
Danielius1922, line 189: Is `a.private.mutex` just protecting the pointer assignment or should it protect the whole `a.private.tlsKeyPair` object? Because when two goroutines call getTLSKeyPair() they get the same pointer outside the lock. If then one changes the object we have a data race.

pkg/security/certManager/general/certManager.go
```diff
@@ -130,124 +143,191 @@ func (a *CertManager) Close() {
...
-		}
-		cert, err := tls.X509KeyPair(tlsCert, tlsKey)
-		if err != nil {
-			return fmt.Errorf("cannot load certificate pair: %w", err)
+func (a *CertManager) loadCertsLocked() (bool, error) {
+	if a.config.KeyFile == "" || a.config.CertFile == "" {
+		return false, nil
+	}
+	keyPath := a.config.KeyFile
+	tlsKey, err := os.ReadFile(keyPath)
```
Danielius1922, line 205: Reading a file under a lock is usually suboptimal, can't we split the reading and setting of the data? Unless some other part of `a` needs to be locked than `a.setTLSKeyPairLocked`, but I see no `a.private` in this function.

pkg/security/certManager/general/certManager_test.go
```diff
@@ -134,3 +136,146 @@ func deleteTmpCertFiles(t *testing.T, cfg general.Config) {
...
+	require.NoError(t, err)
+	defer func() {
+		_ = fileWatcher.Close()
+	}()
+	mng, err := general.New(config, fileWatcher, logger)
+	require.NoError(t, err)
+	defer mng.Close()
+	pool := mng.GetCertificateAuthorities()
+	require.NotNil(t, pool)
+	require.Equal(t, 1, len(pool.Subjects()))
```
Danielius1922: // nolint:staticcheck

tools/nats-server-config-reloader/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+		for sig := range c {
+			log.Printf("Trapped \"%v\" signal\n", sig)
+			switch sig {
+			case syscall.SIGINT:
+				log.Println("Exiting...")
+				os.Exit(0)
+				return
+			case syscall.SIGTERM:
+				r.Stop()
```
Danielius1922: _ = r.Stop()

pkg/security/certManager/general/certManager.go
jkralik: just protecting `a.private.tlsKeyPair` pointer and data in `a.private.tlsKeyPair` is immutable

pkg/security/certManager/general/certManager.go
jkralik: fixed

pkg/security/certManager/general/certManager.go
jkralik: No. Most of time the application uses loaded data in memory. Only when data is expired than it need to load it from the file and all other go-routine must wait until data is loaded (Mutex is much cheaper than reading files from the disc).

tools/nats-server-config-reloader/main.go
jkralik: fixed

pkg/security/certManager/general/certManager_test.go
jkralik: fixed

charts/plgd-hub/values.yaml
```diff
@@ -64,13 +64,13 @@ nats:
...
     noAdvertise: false
-
   leafnodes:
     enabled: false
     noAdvertise: false
-
   natsbox:
     enabled: false
+  reloader:
+    image: ghcr.io/plgd-dev/hub/nats-server-config-reloader:vnext-pr956
```
Danielius1922: Won't this image get deleted when you close this PR?

jkralik: Sure. I want to change it to vnext.

Danielius1922 (approved):

PR #957, opened 2023-06-09
Clean-up untagged packages older than a month

Danielius1922:


jkralik (approved):

jkralik (commented):
We need to be sure that it will work after the images will be removed.

PR #968, opened 2023-06-16
coap-gateway: create swagger documentation

jkralik:


coap-gateway/swagger.yaml
```diff
@@ -0,0 +1,770 @@
...
+          description: OK
+          schema:
+            $ref: "#/definitions/TimeResponse"
+parameters:
+  interface:
+    in: query
+    name: if
+    type: string
+    enum:
+      - oic.if.baseline
```
Danielius1922: only baseline is supported by all that use the param?

ie:
/oic/sec/account
/oic/sec/session
/oic/sec/tokenrefresh
/oic/rd
/api/v1/devices/{deviceId}/{href}

jkralik: Fixed

Danielius1922 (approved):

PR #969, opened 2023-06-22
eventbus: replace resourceId by hrefId in subjects

jkralik:
allows to filter events by hrefs from nats/grpc-gateway/http-gateway


grpc-gateway/subscription/convertToSubjects_test.go
```diff
@@ -4,34 +4,75 @@ import (
...
+			want: []string{
+				isEvents.ToSubject(isEvents.PlgdOwnersOwner+".>", isEvents.WithOwner("")),
+			},
 		},
 		{
-			name: "all",
+			name: "all - DeviceIdFilter *",
+			args: args{
+				req: &pb.SubscribeToEvents_CreateSubscription{
+					DeviceIdFilter: []string{"*", "*"},
```
Danielius1922, line 39: Why `[]string{"*", "*"}` instead of just `[]string{"*"}`? Testing deduplication?

(Same for the following three test cases.)

jkralik: Yes, I want to check if they are deduplicated.

grpc-gateway/subscription/subscription.go
```diff
@@ -139,38 +203,167 @@ func (s *Sub) Close() error {
 	return nil
 }
 
-func New(send SendEventFunc, correlationID string, req *pb.SubscribeToEvents_CreateSubscription) *Sub {
-	bitmask := EventsFilterToBitmask(req.GetEventFilter())
-	filteredResourceIDs := strings.MakeSet()
-	filteredDeviceIDs := strings.MakeSet(req.GetDeviceIdFilter()...)
+func removeDuplicateStrings(s []string) []string {
```
Danielius1922: Can't this be used https://github.com/plgd-dev/hub/blob/main/pkg/strings/slice.go#L45 ?

Danielius1922 (approved):

grpc-gateway/subscription/subscription.go
jkralik: fixed

PR #985, opened 2023-07-03
Upgrade dependencies

Danielius1922:
Submodule:
github.com/googleapis/googleapis 5b2d46f59088b73d9c8db994e49f33d0e2880782

Direct:
github.com/go-co-op/gocron v1.30.1
github.com/lestrrat-go/jwx/v2 v2.0.11
github.com/nats-io/nats.go v1.27.1
github.com/panjf2000/ants/v2 v2.8.0
github.com/plgd-dev/device/v2 v2.2.1-0.20230703110100-49fbc158a560 github.com/plgd-dev/go-coap/v3 v3.1.4-0.20230703104156-ce502465705d github.com/sirupsen/logrus v1.9.3
go.mongodb.org/mongo-driver v1.12.0
golang.org/x/net v0.11.0
golang.org/x/oauth2 v0.9.0
golang.org/x/sync v0.3.0
google.golang.org/api v0.129.0
google.golang.org/genproto/googleapis/rpc v0.0.0-20230629202037-9506855d4529 google.golang.org/grpc v1.56.1
google.golang.org/protobuf v1.31.0

Indirect:
cloud.google.com/go/compute v1.20.1
github.com/envoyproxy/protoc-gen-validate v0.10.1
github.com/googleapis/enterprise-certificate-proxy v0.2.5 github.com/googleapis/gax-go/v2 v2.11.0
github.com/klauspost/compress v1.16.6
github.com/segmentio/asm v1.2.0
go.opentelemetry.io/proto/otlp v0.20.0
golang.org/x/crypto v0.10.0
golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df golang.org/x/sys v0.9.0
golang.org/x/text v0.10.0
google.golang.org/genproto/googleapis/api v0.0.0-20230629202037-9506855d4529


certificate-authority/store/mongodb/signingRecords.go
```diff
@@ -113,7 +116,7 @@ func (s *Store) DeleteSigningRecords(ctx context.Context, owner string, query *s
 	}
 	res, err := s.Collection(signingRecordsCol).DeleteOne(ctx, toSigningRecordsQueryFilter(owner, &q))
 	if err != nil {
-		return -1, fmt.Errorf("cannot remove signing record: %w", err)
+		return -1, pkgErrors.NewError(ErrCannotRemoveSigningRecord, err)
```
jkralik: why you don't use multierror ?

pkg/errors/errors.go
```diff
@@ -0,0 +1,56 @@
+package errors
+
+import (
+	"errors"
+	"fmt"
+)
+
+type wrappedError struct {
```
jkralik: Unwrap is not supported - is seems to me as a duplicate of multierror package.

certificate-authority/store/mongodb/signingRecords.go
Danielius1922: I guess that might work too. I was thinking of `multierror` as multiple errors from the same layer, whilst I just wanted to wrap an existing error. So in my head with multierror I can have multiple unrelated errors whilst this should be a single wrapped error, so when some upper layer wraps it further and then checks it by error.Is(ErrCannotRemoveSigningRecord, err) it should return true. However, that will also holds when `multierror` is used. So it seems that the only difference is the formatted output - "%s: %s" vs "\n\t%s\n\t%s" used by `multierror`, which doesn't warrant a new package.

jkralik (approved):

PR #986, opened 2023-07-04
services: expose build version

jkralik:


bundle/Dockerfile
```diff
@@ -8,60 +8,66 @@ COPY . .
...
+ARG SHORT_COMMIT
+ARG DATE
+ARG VERSION
+ARG RELEASE_URL
+
 #coap-gateway
 ARG service=coap-gateway
 WORKDIR $root_directory/$service
-RUN go build -ldflags "-linkmode external -extldflags -static" -o /go/bin/$service ./cmd/service
+RUN go build -ldflags "-linkmode external -extldflags -static -X github.com/plgd-dev/hub/v2/pkg/build.CommitDate=$COMMIT_DATE -X github.com/plgd-dev/hub/v2/pkg/build.CommitHash=$SHORT_COMMIT -X github.com/plgd-dev/hub/v2/pkg/build.BuildDate=$DATE -X github.com/plgd-dev/hub/v2/pkg/build.Version=$VERSION -X github.com/plgd-dev/hub/v2/pkg/build.ReleaseURL=$RELEASE_URL" -o /go/bin/$service ./cmd/service
```
Danielius1922, line 20: What happends when the `ARG` is not set by the caller of this Dockerfile? 

For example: `COMMIT_DATE` is not be set, so it will call

```
`-X github.com/plgd-dev/hub/v2/pkg/build.CommitDate=`
```

which will write an empty string to `CommitDate` in build.go. 

I guess that shouldn't break anything in the runtime of the services. Or should they always be filled? In that case we should add checks that the values are filled here.

Danielius1922 (approved):

bundle/Dockerfile
jkralik: No, They don't need to be filled. In this case UI doesn't get any information of the build. Same as it is before.

PR #999, opened 2023-07-13
coap-gateway: Correct online_valid_until for tokens without expiration

### LINKED ISSUE #997: Device status is not being updated in case its status changes to offline while hub is down
### Description
In the event of a device being removed from the network while hub is down, when it is back up its state is kept online and is not updated anymore. Looking at the COAP logs no update happens for this particular device. 

### How can it be reproduced?
1. Hub running with device A and B onboard 
2. Shutdown machine running Hub
3. Remove device A from network
4. Bring Hub back up
5. Device A and Device B are reported as being online although only Device B is online

### Environment
- coap gateway:2.7.19
- Deployment option: ( `helm` )

jkralik:
Fixes #997

When a token is assigned an expiration value of 0, it indicates that the token does not have an expiration. In such cases, the connection's online validity should be set according to the configuration in `deviceStatusExpiration.expiresIn`, but only if
`deviceStatusExpiration.enabled` is set to true.

### DIFF
--- a/coap-gateway/service/devicesStatusUpdater.go
+++ b/coap-gateway/service/devicesStatusUpdater.go
@@ -77,7 +77,8 @@ func (u *devicesStatusUpdater) updateOnlineStatus(ctx context.Context, client *s
 		return nil, time.Time{}, err
 	}
 	ctx = kitNetGrpc.CtxWithToken(ctx, authCtx.GetAccessToken())
-	if !u.cfg.Enabled || authCtx.Expire.UnixNano() < validUntil.UnixNano() {
+	// When authCtx.Expire is zero, the token will never expire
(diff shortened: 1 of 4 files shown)


http-gateway/web/package.json
```diff
@@ -43,7 +43,7 @@
     "react-avatar": "^5.0.3",
     "react-bootstrap": "^2.8.0",
     "react-dom": "^18.2.0",
-    "react-error-overlay": "6.0.11",
+    "react-error-overlay": "6.0.9",
```
Danielius1922: This seems unrelated to this PR...?

jkralik: Yes ... patrik has break the build ... will be fixed by him soon

Danielius1922 (approved):

PR #1031, opened 2023-08-11
events: propagate hubID over the events

jkralik:
Assign a `hubID` to every generated event using the identity-store and resource-aggregate mechanisms. This enables the differentiation of hubs in cases where your application spans multiple independent hub deployments.


grpc-gateway/pb/doc.html
```diff
@@ -3694,6 +3694,13 @@ <h3 id="resourceaggregate.pb.EventMetadata">EventMetadata</h3>
                   <td><p>sequence number within the same connection_id; the ResourceChanged event uses the value to skip old events, other event types might not fill the value </p></td>
                 </tr>
               
+                <tr>
+                  <td>hub_id</td>
+                  <td><a href="#string">string</a></td>
+                  <td></td>
+                  <td><p> </p></td>
```
ondrejtomcik: Missing description of the property

grpc-gateway/pb/service.swagger.json
```diff
@@ -1510,6 +1510,9 @@
           "type": "string",
           "format": "uint64",
           "title": "sequence number within the same connection_id; the ResourceChanged event uses the value to skip old events, other event types might not fill the value"
+        },
+        "hubId": {
+          "type": "string"
```
ondrejtomcik: Missing title and format

identity-store/config.yaml
```diff
@@ -55,6 +55,7 @@ clients:
         certFile: "/secrets/public/cert.crt"
         useSystemCAPool: false
   storage:
+    hubID: ""
```
ondrejtomcik: Why is it part of storage and not part of the root service configuration?

identity-store/persistence/persist.go
```diff
@@ -4,6 +4,7 @@ package persistence
 type AuthorizedDevice struct {
 	DeviceID string `db:"device_id"`
 	Owner    string `db:"owner"`
+	HubID    string `db:"hub_id"`
```
ondrejtomcik: I am not aware of the plan to reuse one identity store service for multiple instances of plgd hub. Why do you need to persist the value?
What if the hub id changes? How do you reconcile the database? Should you still see devices - what about querying?

identity-store/service/config.go
```diff
@@ -74,12 +74,16 @@ func (c *EventBusConfig) Validate() error {
...
 type StorageConfig struct {
 	MongoDB mongodb.Config `yaml:"mongoDB" json:"mongoDb"`
+	HubID   string         `yaml:"hubID" json:"hubID"`
 }
 
 func (c *StorageConfig) Validate() error {
 	if err := c.MongoDB.Validate(); err != nil {
 		return fmt.Errorf("mongoDB.%w", err)
 	}
+	if c.HubID == "" {
```
ondrejtomcik: We need to validate that the hub ID is a valid uuid.

resource-aggregate/config.yaml
```diff
@@ -62,6 +62,7 @@ clients:
     snapshotThreshold: 16
     # limits number of try to store event
     occMaxRetry: 8
+    hubID: ""
```
ondrejtomcik: Please unify the location of this configuration option. In the identity store, you have it in the storage. Here you have it under the event store. Please consider moving it to the root.

resource-aggregate/service/config.go
```diff
@@ -81,6 +82,9 @@ func (c *EventStoreConfig) Validate() error {
 	if c.ConcurrencyExceptionMaxRetry <= 0 {
 		return fmt.Errorf("occMaxRetry('%v')", c.ConcurrencyExceptionMaxRetry)
 	}
+	if c.HubID == "" {
```
ondrejtomcik: Don't forget to validate the format.

resource-aggregate/service/aggregate.go
```diff
@@ -17,25 +17,27 @@ type LogPublishErrFunc func(err error)
...
 }
 
 func DeviceMetadataFactoryModel(context.Context) (cqrsAggregate.AggregateModel, error) {
-	return raEvents.NewDeviceMetadataSnapshotTaken(), nil
+	return events.NewDeviceMetadataSnapshotTaken(), nil
 }
 
 // NewAggregate creates new resource aggreate - it must be created for every run command.
-func NewAggregate(resourceID *commands.ResourceId, snapshotThreshold int, eventstore EventStore, factoryModel cqrsAggregate.FactoryModelFunc, retry cqrsAggregate.RetryFunc) (*Aggregate, error) {
+func NewAggregate(resourceID *commands.ResourceId, snapshotThreshold int, hubID string, eventstore EventStore, factoryModel cqrsAggregate.FactoryModelFunc, retry cqrsAggregate.RetryFunc) (*Aggregate, error) {
```
ondrejtomcik, line 37: When creating an aggregate, you always pass the hubID from the configuration. When you then load events, you do not care if it is the matching one or not. And when handling command and publishing the event, you always take the hub id from the context -> aggregate -> configuration, while in identity store you take the one stored into the db. You use two different approaches. Please come up with a consistent solution - e.g. updating the identity store.

ondrejtomcik (changes requested):

grpc-gateway/pb/doc.html
jkralik: This is generated doc. I don't want to spend time with the generator.

identity-store/config.yaml
jkralik: moved to root

grpc-gateway/pb/service.swagger.json
jkralik: fixed

identity-store/persistence/persist.go
jkralik: removed from persistent storage

identity-store/service/config.go
jkralik: fixed

resource-aggregate/config.yaml
jkralik: moved to root

resource-aggregate/service/config.go
jkralik: fixed

resource-aggregate/service/aggregate.go
jkralik: Fixed -> removed from IS storage

charts/plgd-hub/README.md
```diff
@@ -85,6 +85,7 @@ global:
 | certificateauthority.extraVolumes | string | `nil` | Optional extra volumes |
 | certificateauthority.fullnameOverride | string | `nil` | Full name to override |
 | certificateauthority.httpPort | int | `9101` |  |
+| certificateauthority.hubId | string | `nil` | Hub ID. Can be override via global.hubId or coapgateway.hubId |
```
Danielius1922: can be overriden

same typo on lines 422 and 570

grpc-gateway/pb/devices.proto
```diff
@@ -112,13 +113,13 @@ message SubscribeToEvents {
 message Event {
   message DeviceRegistered {
     repeated string device_ids = 1;
-
+    identitystore.pb.EventMetadata event_metadata = 5; // provides metadata of event
```
Danielius1922: why 5 and not 2? are 2-4 reserved?

jkralik: This is a issue ... `DeviceRegistered/DeviceDeregistered` I want replace them by events from identity store. But we need to introduce new type ...

charts/plgd-hub/README.md
jkralik: fixed

Danielius1922 (approved):

charts/plgd-hub/values.yaml
```diff
@@ -432,6 +432,8 @@ resourcedirectory:
     mountPath: /config
   # -- Service and POD port
   port: 9100
+  # -- Hub ID. Can be overriden via global.hubId
```
ondrejtomcik: Can be overriden or this can override the global.hubId ?

ondrejtomcik (approved):

charts/plgd-hub/values.yaml
jkralik: fixed

PR #1049, opened 2023-09-07
certificate-authority: allow sets external caPool to signer for k8s

jkralik:
The Certificate Authority (`signer.certFile`) expects a complete chain of certificates, starting from the intermediate sign cert up to the root cert. However, in Kubernetes (k8s), the root cert is not included in the `tls.crt` file; it is stored separately as `ca.crt`. Therefore, this commit allows you to specify the CA file as well via the `signer.caPool`. This change enables the resolution of the full certificate chain, which is used during certificate signing.

### DIFF
--- a/certificate-authority/service/grpc/server.go
+++ b/certificate-authority/service/grpc/server.go
@@ -1,60 +1,113 @@
 package grpc
 
 import (
-	"context"
-	"crypto"
-	"crypto/x509"
-	"time"
+	"bytes"
+	"fmt"
 
-	"github.com/karrick/tparse/v2"
 	"github.com/plgd-dev/hub/v2/certificate-authority/pb"
 	"github.com/plgd-dev/hub/v2/certificate-authority/store/mongodb"
+	"github.com/plgd-dev/hub/v2/pkg/fn"
+	"github.com/plgd-dev/hub/v2/pkg/fsnotify"
 	"github.com/plgd-dev/hub/v2/pkg/log"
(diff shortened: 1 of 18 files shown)


certificate-authority/service/grpc/server.go
```diff
@@ -1,60 +1,112 @@
...
-	privateKey, err := security.LoadX509PrivateKey(signerConfig.KeyFile)
+
+	var removeFilesOnError fn.FuncList
+	for _, ca := range signerConfig.caPoolArray {
+		if err := fileWatcher.Add(ca); err != nil {
+			removeFilesOnError.Execute()
+			return nil, fmt.Errorf("cannot watch CAPool(%v): %w", ca, err)
+		}
+		removeFilesOnError.AddFunc(func() {
+			_ = fileWatcher.Remove(ca)
```
Danielius1922: if I recall corectly Go captures variables as references in closures, so all functions added here will have the same `ca`.

Assigning to helper variable should work here:

```Go
for _, ca := range signerConfig.caPoolArray {
  ...
  caToRemove := ca
  removeFilesOnError.AddFunc(func() {
    _ = fileWatcher.Remove(caToRemove)
  }
}

```

jkralik: Good point - fixed

Danielius1922 (approved):

ondrejtomcik (approved):

PR #1058, opened 2023-09-21
Service Monitoring: Establish Association between Service and Device

jkralik:
When a service has been terminated, and devices have not transitioned
to an "offline" status, they are incorrectly reported as "online." By
introducing this association, all devices belonging to the service
can now be updated to the "offline" state automatically.

This association is stored in the database, similar to device
resources. In this case, the hub  represents a device, and the content
is a special resource containing all services, each with a validity
period. This means that each service must update its validity into
the future. If a service goes down and its validity expires, any other
 service that updates its validity will trigger a check to see if any
service is expired. If any service is indeed expired, all associated
devices will also be updated to the "offline" state.

For more information, visit the [link](https://docs.plgd.dev/docs/features/monitoring-and-diagnostics/monitoring-device-connectivity/).


resource-aggregate/pb/commands.proto
```diff
@@ -657,3 +662,19 @@ message DeleteDevicesResponse {
     repeated string device_ids = 1;
     AuditContext audit_context = 100;
 }
+
+message ServiceStatus {
+    ServiceID service_id = 1;
```
jkralik: enum - coap-gw, c2c, grpc-gw, unknown

resource-aggregate/pb/commands.proto
```diff
@@ -657,3 +662,19 @@ message DeleteDevicesResponse {
...
+    string instance_id = 2; // generated unique id during start the service
```
jkralik: uniqe

resource-aggregate/pb/resources.proto
```diff
@@ -7,6 +7,7 @@ option go_package = "github.com/plgd-dev/hub/v2/resource-aggregate/commands;comm
 message AuditContext {
   string user_id = 1;
   string correlation_id = 2;
+  string owner = 3;
```
jkralik, line 10: remove owner - use owners from IS

resource-aggregate/config.yaml
```diff
@@ -1,4 +1,7 @@
 hubID: ""
+serviceStatus:
+  owner: "admin"
+  userID: "admin"
```
jkralik: hardcode - create name

resource-aggregate/config.yaml
```diff
@@ -1,4 +1,7 @@
 hubID: ""
+serviceStatus:
+  owner: "admin"
```
jkralik: remove

jkralik: done

resource-aggregate/config.yaml
jkralik: done

resource-aggregate/pb/resources.proto
jkralik: @ondrejtomcik need to be decided

resource-aggregate/pb/commands.proto
jkralik: `ServiceID service_id = 1;` - removed
string instance_id = 2;  -> string id = 1;

resource-aggregate/pb/commands.proto
jkralik: removed

resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
```diff
@@ -270,6 +277,18 @@ func makeDBETag(etag *eventstore.ETagData) bson.M {
 	}
 }
 
+func tryToSetServiceID(doc bson.M, events []eventstore.Event) bson.M {
+	var serviceID string
+	var ok bool
+	for _, e := range events {
+		serviceID, ok = e.ServiceID()
```
Danielius1922: This code is intended to get the serviceID of the last event? If so I would expect something like this instead of a for cycle:

```go
e, ok := events[len(events)-1].ServiceID()
```

resource-aggregate/service/service.go
```diff
@@ -17,17 +17,17 @@ import (
...
-	cqrsEventStore "github.com/plgd-dev/hub/v2/resource-aggregate/cqrs/eventstore"
-	cqrsMaintenance "github.com/plgd-dev/hub/v2/resource-aggregate/cqrs/eventstore/maintenance"
-	mongodb "github.com/plgd-dev/hub/v2/resource-aggregate/cqrs/eventstore/mongodb"
+	"github.com/plgd-dev/hub/v2/resource-aggregate/cqrs/eventstore/mongodb"
 	"github.com/plgd-dev/hub/v2/resource-aggregate/cqrs/utils"
 	"go.opentelemetry.io/otel/trace"
 )
 
+/*
 type EventStore interface {
```
Danielius1922: rm?

resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
jkralik: fixed

resource-aggregate/service/service.go
jkralik: removed

resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
```diff
@@ -270,6 +277,16 @@ func makeDBETag(etag *eventstore.ETagData) bson.M {
 	}
 }
 
+func tryToSetServiceID(doc bson.M, events []eventstore.Event) bson.M {
+	for _, e := range events {
+		serviceID, ok := e.ServiceID()
+		if ok {
+			doc[serviceIDKey] = serviceID
```
Danielius1922, line 284: What's happening here? Are the events somehow ordered and we know that a later serviceID is better than a serviceID from an earlier event? Why prefer the last match instead of for example the first match?

resource-aggregate/events/deviceMetadataSnapshotTaken_test.go
```diff
@@ -168,7 +162,7 @@ func TestDeviceMetadataSnapshotTakenHandleCommand(t *testing.T) {
 			name: "online,online,offline",
 			cmds: []cmd{
 				{
-					ctx: events.CtxWithHubID(grpc.CtxWithIncomingToken(context.Background(), jwtWithSubUserID), hubID),
+					ctx: context.Background(),
```
Danielius1922: I think you can remove the `ctx` member from the `cmd` struct and just use `context.Background` in the call.

resource-aggregate/events/servicesMetadataUpdated.go
```diff
@@ -0,0 +1,93 @@
...
+		d.GetEventMetadata() != nil
+}
+
+func (s *ServicesStatus) CopyData(s1 *ServicesStatus) {
+	s.Online = s1.GetOnline()
+	s.Offline = s1.GetOffline()
+}
+
+func (s *ServicesStatus) Equal(upd *ServicesStatus) bool {
+	if len(s.GetOnline()) != len(upd.GetOnline()) {
```
Danielius1922: What about the GetOffline array? That doesn't need to be compared?

grpc-gateway/service/subscribeToEvents_test.go
```diff
@@ -944,3 +947,148 @@ func TestRequestHandlerIssue270(t *testing.T) {
...
+		}, ev, "")
+		if wantBreak {
+			return firstCoapGWInstanceID
+		}
+	}
+}
+
+func TestCoAPGatewayServiceStatus(t *testing.T) {
+	deviceID := test.MustFindDeviceByName(test.TestDeviceName)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3600)
```
Danielius1922: 1 hour?

resource-aggregate/cqrs/eventstore/mongodb/eventstore.go
jkralik: the latest event is the current state -> so if you have in first event with coap-gw-1 and second event contains the coap-gw-2 then the coap-gw-2 is the current coap-gw.

resource-aggregate/service/cancelDeviceMetadataUpdates.go
```diff
@@ -64,7 +64,7 @@ func (r RequestHandler) CancelPendingMetadataUpdates(ctx context.Context, reques
 	}
 
 	return &commands.CancelPendingMetadataUpdatesResponse{
-		AuditContext:   commands.NewAuditContext(owner, ""),
+		AuditContext:   commands.NewAuditContext(owner, "", owner),
```
Danielius1922: Usually I see `user` as the first paramete to `NewAuditContext`, so is using `owner` for both (instead of using `userID` as the first param) correct here?

grpc-gateway/service/subscribeToEvents_test.go
jkralik: fixed

resource-aggregate/service/cancelDeviceMetadataUpdates.go
jkralik: fixed

resource-aggregate/events/servicesMetadataUpdated.go
jkralik: fixed

resource-aggregate/events/deviceMetadataSnapshotTaken_test.go
jkralik: removed

resource-aggregate/events/servicesMetadataSnapshotTaken.go
```diff
@@ -0,0 +1,328 @@
...
+		offline[v.GetId()] = v
+		delete(online, v.GetId())
+	}
+	// check if there is no service which is online and offline at the same time
+	for key := range online {
+		if _, ok := offline[key]; ok {
+			return false, fmt.Errorf("invalid status: service %v is online and offline", key)
+		}
+	}
+	for key := range offline {
```
Danielius1922: Don't need this second iteration, if a key is present in both `online` and `offline` maps then the first iteration will find it.

resource-aggregate/events/servicesMetadataSnapshotTaken.go
```diff
@@ -0,0 +1,328 @@
...
+	// When the processing time exceeds the timeToLive, we must extend the timeToLive.
+	// During processing, several tasks are performed: receiving requests, loading data from the database, and processing requests.
+	// Writing to the database is an expensive operation, so if the processing time is more than 10 times the timeToLive, we should extend it.
+	// During idle times, the processing time is typically around 10 milliseconds.
+	// Under heavy load, the processing time can increase to approximately 5 minutes.
+	timeToLive := time.Duration(req.GetStatus().GetTimeToLive())
+	if req.GetStatus().GetTimestamp() >= 0 {
+		processingTime := time.Since(pkgTime.Unix(0, req.GetStatus().GetTimestamp()))
+		if processingTime*writeCostAgainstRead > timeToLive {
+			timeToLive = processingTime * writeCostAgainstRead
```
Danielius1922: So under heavy load this could be 50+ minutes, does the next metadata update override the expiration time? Or is it possible that a service will remain valid even though the heavy load is long gone?

resource-aggregate/events/servicesMetadataSnapshotTaken.go
```diff
@@ -0,0 +1,328 @@
...
+
+	removed := make([]*ServicesStatus_Status, 0, len(req.Status))
+	for _, v := range req.Status {
+		key := v.GetId()
+		if _, ok := offline[key]; ok {
+			delete(offline, key)
+			removed = append(removed, v)
+		}
+	}
+	if len(removed) == 0 {
```
Danielius1922: We can save up a little bit of memory, since we only need to know whether at least one element was removed, you only require a single `bool` for that

resource-aggregate/events/servicesMetadataSnapshotTaken.go
jkralik: Yes it will be updated. CoAP-gw tries to update in 1/3 of `timeToLive` -> so it will be updated after 16minutes back to eg 30sec if RA and mongo are idle again.

Danielius1922 (approved):

resource-aggregate/events/servicesMetadataSnapshotTaken.go
jkralik: fixed

resource-aggregate/events/servicesMetadataSnapshotTaken.go
jkralik: removed

charts/plgd-hub/values.yaml
```diff
@@ -881,6 +881,10 @@ coapgateway:
   port: 5684
   # -- Hub ID. Overrides the global.hubId
   hubId:
+  # -- Service status section
+  serviceStatus:
+    # -- Time to update the service status of the service. Must be greater than 1s.
```
ondrejtomcik: `# Specifies validity of the presence record created by the gateway.`

resource-aggregate/pb/commands.proto
```diff
@@ -657,3 +661,20 @@ message DeleteDevicesResponse {
...
+
+message UpdateServiceMetadataRequest {
+    oneof update {
+        // Add or update service in the online services list and move online expired services to offline services list
+        ServiceStatus status = 1;
+    }
+}
+
+message UpdateServiceMetadataResponse {
+    int64 online_valid_until = 1;  // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano) when service is considered as expired. 0 means forever.
```
ondrejtomcik: `heartbeat_valid_until`

resource-aggregate/pb/commands.proto
```diff
@@ -423,6 +423,9 @@ message ConfirmResourceCreateResponse {
 //*******************************************************************************************************************************************************
 
 message Connection {
+    message Service {
+        string id = 1; // generated unique id during start the service
```
ondrejtomcik: Why not directly `service_id` as part of connection?

resource-aggregate/pb/commands.proto
```diff
@@ -657,3 +661,20 @@ message DeleteDevicesResponse {
...
+message ServiceStatus {
+    string id = 1; // generated unique id during start the service
+    int64 time_to_live = 2; // validity in nanoseconds. minimal value is 1000000000 (1s).
+    int64 timestamp = 3; // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano)
+}
+
+message UpdateServiceMetadataRequest {
+    oneof update {
+        // Add or update service in the online services list and move online expired services to offline services list
+        ServiceStatus status = 1;
```
ondrejtomcik: `ServiceHeartbeat`? Status is a result of processing

resource-aggregate/pb/commands.proto
```diff
@@ -657,3 +661,20 @@ message DeleteDevicesResponse {
     repeated string device_ids = 1;
     AuditContext audit_context = 100;
 }
+
+message ServiceStatus {
+    string id = 1; // generated unique id during start the service
```
ondrejtomcik: Is it service id or status id? Isn't this part of the connection already?

charts/plgd-hub/values.yaml
```diff
@@ -881,6 +881,10 @@ coapgateway:
   port: 5684
   # -- Hub ID. Overrides the global.hubId
   hubId:
+  # -- Service status section
+  serviceStatus:
```
ondrejtomcik: What do you think about `serviceHeartbeat` ?

ondrejtomcik: on a root level

resource-aggregate/pb/commands.proto
ondrejtomcik: Rename to service_id

resource-aggregate/pb/events.proto
```diff
@@ -205,3 +205,28 @@ message DeviceMetadataSnapshotTaken {
...
 }
+
+
+message ServicesHeartbeat {
+    message Heartbeat {
+        string service_id = 1; // generated unique id during start the service
+        int64 heartbeat_valid_until = 2; // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano) when service heartbeat is considered as expired.
+    }
+    repeated Heartbeat online = 1; // services which are online
+    repeated Heartbeat offline = 2; // services which are offline
```
ondrejtomcik: heartbeat can be expired or valid, not online or offline

resource-aggregate/pb/events.proto
```diff
@@ -205,3 +205,28 @@ message DeviceMetadataSnapshotTaken {
...
+}
+
+message ServicesMetadataUpdated {
+    ServicesHeartbeat heartbeat  = 1;
```
ondrejtomcik: heartbeats

resource-aggregate/pb/events.proto
```diff
@@ -205,3 +205,28 @@ message DeviceMetadataSnapshotTaken {
     // Open telemetry data propagated to asynchronous events
    // map<string,string> open_telemetry_carrier = 100;
 }
+
+
+message ServicesHeartbeat {
+    message Heartbeat {
+        string service_id = 1; // generated unique id during start the service
+        int64 heartbeat_valid_until = 2; // unix timestamp in nanoseconds (https://golang.org/pkg/time/#Time.UnixNano) when service heartbeat is considered as expired.
```
ondrejtomcik: as it's part of Heartbeat, valid_until is sufficient

resource-aggregate/pb/events.proto
ondrejtomcik: services which heartbeat is already expired
services which heartbeat is still valid

--> doc

ondrejtomcik (approved):

PR #1076, opened 2023-10-03
Extend the gRPC/HTTP Gateway API to support ETag in the resourceIdFilter

jkralik:
This update enables the API to handle ETag values in the resourceIdFilter.
When using gRPC or HTTP streaming, the response for an unmodified resource
will have `data.status` set to `NOT_MODIFIED`, and `data.content.data`
will be empty. Additionally, for HTTP GET requests to
`/api/v1/devices/{deviceId}/resources/{resourceHref}` with an ETag in
the header, the response will return an HTTP status code indicating that
the resource has not been modified.


grpc-gateway/pb/devices.proto
```diff
@@ -37,16 +38,24 @@ message GetResourceFromDeviceRequest {
...
+
 message GetResourcesRequest {
-  repeated string resource_id_filter = 1; // format {deviceID}{href}. eg "ae424c58-e517-4494-6de7-583536c48213/oic/d"
-  repeated string device_id_filter = 2;
-  repeated string type_filter = 3;
+  repeated string http_resource_id_filter = 1 [ deprecated = true ];  // Format: {deviceID}{href}(?etag=abc), e.g., "ae424c58-e517-4494-6de7-583536c48213/oic/d?etag=abc"
+  repeated string device_id_filter = 2; // Filter devices by deviceID
+  repeated string type_filter = 3; // Filter devices by resource types in the oic/d resource
+
+  repeated ResourceIdFilter resource_id_filter = 5; // New resource ID filter. For HTTP requests, use it multiple times as a query parameter like "resourceIdExtendedFilter=protojson.Marshal(ResourceIdFilter)"
```
Danielius1922: Is `4` reserved for different use or why is it skipped and `5` is used instead?

grpc-gateway/pb/events.proto
```diff
@@ -3,13 +3,14 @@ syntax = "proto3";
...
-
+import "grpc-gateway/pb/devices.proto";
 option go_package = "github.com/plgd-dev/hub/v2/grpc-gateway/pb;pb";
 
 message GetEventsRequest {
 	repeated string device_id_filter = 1;
 	// format {deviceID}{href}. eg "ae424c58-e517-4494-6de7-583536c48213/oic/d"
-	repeated string resource_id_filter = 2;
+	repeated string http_resource_id_filter = 2 [deprecated = true];
+	repeated ResourceIdFilter resource_id_filter = 4;
```
Danielius1922: `=3` ?

grpc-gateway/pb/devices.go
```diff
@@ -64,3 +69,28 @@ func (e *Event_ResourceCreated) GetResourceId() *commands.ResourceId {
...
+		return sb.String()
+	}
+	for i, etag := range f.GetEtag() {
+		if i == 0 {
+			sb.WriteString("?")
+		} else {
+			sb.WriteString("&")
+		}
+		sb.WriteString("etag=")
+		sb.WriteString(base64.StdEncoding.EncodeToString(etag))
```
Danielius1922, line 93: Shouldn't `base64.RawURLEncoding` be used to avoid the use of URL problematic characters? If it's never used in an URI then it's OK.

coap-gateway/coapconv/coapconv.go
```diff
@@ -34,7 +35,7 @@ func StatusToCoapCode(status commands.Status, operation Operation) codes.Code {
 		}
 	case commands.Status_CREATED:
 		return codes.Created
-	case commands.Status_ACCEPTED:
+	case commands.Status_NOT_MODIFIED:
```
Danielius1922, line 39: if `StatusToCoapCode` is called with  `commands.Status_ACCEPTED` it will now return `codes.BadRequest`. Is that correct behavior?

grpc-gateway/pb/devices.proto
jkralik: fixed

grpc-gateway/pb/events.proto
jkralik: 3 has been used by timestamp

coap-gateway/coapconv/coapconv.go
jkralik: fixed

grpc-gateway/pb/devices.go
jkralik: The final string is encoded by `base64.RawURLEncoding` -> because it is will be `{deviceID}{href}?etag={base64}`
and the main reason is that the protojson.Encoder uses `base64.StdEncoding.EncodeToString` for []byte types. So user can just copy the etag string.

grpc-gateway/pb/events.proto
Danielius1922: ah sorry, I'm blind

http-gateway/service/getResource.go
```diff
@@ -1,29 +1,63 @@
...
+	if etagStr != "" {
+		if etag, err := base64.StdEncoding.DecodeString(etagStr); err == nil {
+			return [][]byte{etag}
+		}
+	}
+	etagQ := r.URL.Query()[uri.ETagQueryKey]
+	if len(etagQ) == 0 {
+		return nil
+	}
+	etags := make([][]byte, 0, len(etagStr))
```
Danielius1922: `etagStr` is equal to `""` in this code branch (see line 31)

EDIT: I think `etagQ` should have been used here

jkralik: You are right :)

Danielius1922 (approved):

ondrejtomcik (approved):

PR #1078, opened 2023-10-11
Configs: Add support for the data URI scheme property.

jkralik:
In certain cases, it is necessary to store property values as raw data. This is applicable to certFile, keyFile, caPool, and clientSecretFile, which now also support the data URI scheme. In most instances, the data must be encoded using standard Base64 and stored in the data scheme format: `data:;base64,<BASE64_DATA>`.


pkg/config/property/urischeme/urischeme.go
```diff
@@ -0,0 +1,113 @@
...
+
+// IsFile returns true if the URIScheme is a file scheme.
+func (p URIScheme) IsFile() bool {
+	if len(p) == 0 {
+		return false
+	}
+	if p.IsData() {
+		return false
+	}
+	if strings.HasPrefix(string(p), "file:///") {
```
Danielius1922, line 55: relative paths are not supported? `file://path/to/file` won't match this prefix but will match `anyScheme`.

Danielius1922: OK, apparently it's correct, relative paths are not supported by the URI scheme strings

pkg/config/property/urischeme/urischeme_test.go
```diff
@@ -0,0 +1,202 @@
...
+func TestToURISchemeArray(t *testing.T) {
+	tests := []struct {
+		name string
+		v    []string
+		want []urischeme.URIScheme
+	}{
+		{
+			name: "empty slice",
+			v:    []string{},
+			want: []urischeme.URIScheme{},
```
Danielius1922: according to the test failures this should be:

```go
want: nil,
```

certificate-authority/service/grpc/server.go
```diff
@@ -44,43 +44,58 @@ func NewCertificateAuthorityServer(ownerClaim string, hubID string, signerConfig
...
-	if err := fileWatcher.Add(signerConfig.KeyFile); err != nil {
-		removeFilesOnError.Execute()
-		return nil, fmt.Errorf("cannot watch KeyFile(%v): %w", signerConfig.CertFile, err)
+
+	if signerConfig.KeyFile.IsFile() {
+		if err := fileWatcher.Add(signerConfig.KeyFile.FilePath()); err != nil {
+			removeFilesOnError.Execute()
+			return nil, fmt.Errorf("cannot watch KeyFile(%v): %w", signerConfig.KeyFile, err)
+		}
+		s.onFileChangeFunc = s.onFileChange
```
Danielius1922: The `onFileChangeFunc` is only relevant when `KeyFile` changes?

Danielius1922 (approved):

PR #1086, opened 2023-10-19
resource-directory: Fix "getEvents" to return empty data when filters are not satisfied

jkralik:


http-gateway/service/getEvents_test.go
```diff
@@ -116,8 +116,7 @@ func TestRequestHandlerGetEvents(t *testing.T) {
 				accept:   uri.ApplicationProtoJsonContentType,
 				deviceID: "test",
 			},
-			wantErr:      true,
-			wantHTTPCode: http.StatusInternalServerError,
+			wantHTTPCode: http.StatusOK,
```
Danielius1922, line 119: Is this correct behavior? `StatusOK` when invalid deviceID is used?

jkralik: GetResources/GetDevices returns empty values when filter is not satisfied.

jkralik: It is similar as filter in gmail -> you don't get error but just empty body.

Danielius1922 (approved):

PR #1088, opened 2023-10-20
Experimental: cqldb: allow to use cqldb instead of mongodb

jkralik:


bundle/README.md
```diff
@@ -53,6 +53,9 @@ Image can be configured via environment variables as argument `-e ENV=VALUE` of
 | `OPEN_TELEMETRY_EXPORTER_CERT_FILE` | string | File path to certificate in PEM format | `"/certs/otel/cert.crt"` |
 | `OPEN_TELEMETRY_EXPORTER_KEY_FILE` | string | File path to private key in PEM format | `"/certs/otel/cert.key"` |
 | `OPEN_TELEMETRY_EXPORTER_CA_POOL` | string | File path to the root certificate in PEM format which might contain multiple certificates in a single file | `"/certs/otel/rootca.crt"` |
+| `DATABASE` | string | Database type. Supported values are: [ "mongo", "scylla" ] | `"mongo"` |
+| `SCYLLA_SMP` | uint16 | Associate number of cpu's with the Scylla DB | `1` |
+| `SCYLLA_DEVELOPER_MODE=1` | bool | Enable the Scylla developer mode | `1` |
```
Danielius1922: | `SCYLLA_DEVELOPER_MODE=1` | bool | Enable the Scylla developer mode | `1` |

| `SCYLLA_DEVELOPER_MODE` | bool | Enable the Scylla developer mode | `true` |

certificate-authority/pb/README.md
```diff
@@ -182,19 +182,19 @@
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| id | [string](#string) |  | Registration id, calculated from the certificate public key info &#43; common name. SHA256 - Must be unique in the database and for one private key there could be multiple certificates.
+| id | [string](#string) |  | Registration id, calculated from the certificate by in uuid.NewSHA1(uuid.NameSpaceX500, common_name &#43; uuid.NewSHA1(uuid.NameSpaceX500, publicKeyRaw)) - Must be unique in the database and for one private key there could be multiple certificates.
```
Danielius1922: Čo chce povedať tá prvá veta?

.vscode/settings.json
```diff
@@ -34,6 +34,7 @@
         "TEST_GRPC_GATEWAY_LOG_DUMP_BODY":  "false",
         "TEST_IDENTITY_STORE_LOG_LEVEL": "info",
         "TEST_IDENTITY_STORE_LOG_DUMP_BODY":  "false",
+        "TEST_DATABASE": "cqlDB",
```
Danielius1922: I would keep mongo as the default when running from VSC.

grpc-gateway/service/createResource_test.go
```diff
@@ -146,7 +148,11 @@ func TestRequestHandlerCreateResource(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), config.TEST_TIMEOUT)
 	defer cancel()
 
-	tearDown := service.SetUp(ctx, t)
+	coapgwCfg := coapTest.MakeConfig(t)
+	coapgwCfg.Log.DumpBody = true
+	coapgwCfg.Log.Level = log.DebugLevel
```
Danielius1922: Any reason to keep debug logs for this test?

bundle/run.sh
```diff
@@ -46,6 +46,7 @@ export OAUTH_ACCESS_TOKEN_KEY_PATH=${OAUTH_KEYS_PATH}/access-token.pem
 export OAUTH_DEVICE_SECRET_PATH=${OAUTH_SECRETS_PATH}/device.secret
 
 #ENDPOINTS
+export SCYLLA_HOST="localhost:29142"
```
Danielius1922: Add `SCYLLA_PORT=...` somewhere near the definitions of `MONGO_PORT` and `NATS_PORT`

identity-store/service/service.go
```diff
@@ -60,19 +80,13 @@ func NewServer(ctx context.Context, cfg Config, fileWatcher *fsnotify.Watcher, l
...
-	if err != nil {
-		return nil, fmt.Errorf("cannot create cert manager %w", err)
-	}
-	grpcServer.AddCloseFunc(certManager.Close)
-
-	persistence, err := mongodb.NewStore(ctx, cfg.Clients.Storage.MongoDB, certManager.GetTLSConfig(), tracerProvider)
+	persistence, err := newPersistence(ctx, cfg.Clients.Storage, fileWatcher, logger, tracerProvider)
 	if err != nil {
-		return nil, fmt.Errorf("cannot create connector to mongo: %w", err)
+		return nil, fmt.Errorf("cannot create connector to persistence storage: %w", err)
```
Danielius1922: persistent storage

identity-store/service/service.go
```diff
@@ -60,19 +80,13 @@ func NewServer(ctx context.Context, cfg Config, fileWatcher *fsnotify.Watcher, l
...
 	}
 	grpcServer.AddCloseFunc(func() {
 		if err := persistence.Close(ctx); err != nil {
-			log.Debugf("failed to close mongodb connector: %w", err)
+			log.Debugf("failed to persistence storage: %w", err)
```
Danielius1922: failed to close persistent storage

bundle/Dockerfile
```diff
@@ -130,10 +131,16 @@ COPY --from=build /go/src/github.com/plgd-dev/hub/cloud2cloud-connector/config.y
 COPY --from=build /go/src/github.com/plgd-dev/hub/bundle/run.sh /usr/local/bin/run.sh
 COPY --from=build /go/src/github.com/plgd-dev/hub/bundle/nginx /nginx
 
+# install scylla
+RUN curl -sSf get.scylladb.com/server | sudo bash -s -- --scylla-version 5.2
```
Danielius1922, line 135: Do we want to hard code to version `5.2`?

jkralik: Yes. We want to update scylla manually - we don't know if the newest version is compatible with the bundle scylla configuration.

.vscode/settings.json
jkralik: fixed

grpc-gateway/service/createResource_test.go
jkralik: removed

bundle/run.sh
jkralik: fixed

identity-store/service/service.go
jkralik: fixed

identity-store/service/service.go
jkralik: fixed

certificate-authority/pb/README.md
jkralik: fixed

bundle/README.md
jkralik: fixed

grpc-gateway/service/updateResource_test.go
```diff
@@ -286,7 +286,6 @@ func TestRequestHandlerRunMultipleUpdateResource(t *testing.T) {
 	defer cancel()
 
 	raCfg := raTest.MakeConfig(t)
-	raCfg.Clients.Eventstore.SnapshotThreshold = 5
```
Danielius1922, line 289: What changed with the SnapShot behavior? (I suspect it will be simpler to explain this on SM tomorrow, so just take this as a reminder).

certificate-authority/store/cqldb/signingRecords.go
```diff
@@ -0,0 +1,462 @@
...
+			count++
+		}
+	} else if !errors.Is(err, gocql.ErrNotFound) {
+		return 0, err
+	}
+	return count, nil
+}
+
+func (s *Store) DeleteNonDeviceExpiredRecords(_ context.Context, _ time.Time) (int64, error) {
+	// Data
```
Danielius1922: Data? I recall some comment that `cqldb` automatically deletes expired records, I would copy that comment here also.

certificate-authority/store/cqldb/signingRecords_test.go
```diff
@@ -0,0 +1,579 @@
...
+	lcs := upds
+
+	type args struct {
+		owner string
+		query *store.SigningRecordsQuery
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
```
Danielius1922, line 432: unused, seems that there is no way of triggering error with just a query. You could create a handler that returns an error, but `signingRecords.go` has already 90+% code coverage, so I think we can just remove this member

certificate-authority/store/cqldb/store.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/plgd-dev/hub/v2/pkg/cqldb"
+	"github.com/plgd-dev/hub/v2/pkg/fsnotify"
+	"github.com/plgd-dev/hub/v2/pkg/log"
+	"github.com/plgd-dev/hub/v2/pkg/security/certManager/client"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// Document
+const (
+	// cqldbdb has all keys in in lowercase
```
Danielius1922: typo: `in in` -> `in`

certificate-authority/store/cqldb/store.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func New(ctx context.Context, config *Config, fileWatcher *fsnotify.Watcher, logger log.Logger, tracerProvider trace.TracerProvider) (*Store, error) {
+	certManager, err := client.New(config.Embedded.TLS, fileWatcher, logger)
+	if err != nil {
+		return nil, fmt.Errorf("could not create cert manager: %w", err)
+	}
+	cqldbClient, err := cqldb.New(ctx, config.Embedded, certManager.GetTLSConfig(), logger, tracerProvider)
+	if err != nil {
+		return nil, err
```
Danielius1922, line 51: `certManager.Close` is not needed before returning?

same for line 56

certificate-authority/store/cqldb/store.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Clear documents in collections, but don't drop the database or the collections
+func (s *Store) ClearTable(ctx context.Context) error {
+	return s.client.Session().Query("truncate " + s.Table() + ";").WithContext(ctx).Exec()
+}
+
+// Close closes the database session.
+func (s *Store) Close(_ context.Context) error {
+	s.client.Close()
+	for _, f := range s.closeFunc {
```
Danielius1922: you can save two lines here if you use our `FuncList` type! 😺

identity-store/persistence/cqldb/persist.go
```diff
@@ -0,0 +1,227 @@
...
+	b.WriteString(cqldb.SelectCommand + " ")
+	b.WriteString(ownerKey)
+	b.WriteString(",")
+	b.WriteString(deviceIDKey)
+	b.WriteString(" " + cqldb.FromClause + " ")
+	b.WriteString(p.table)
+	b.WriteString(" " + cqldb.WhereClause + " ")
+	b.WriteString(deviceIDKey)
+	b.WriteString("=")
+	b.WriteString(deviceID)
```
Danielius1922, line 83: SC reports duplication for `Retrieve` and `RetrieveByDevice`. It can be solved by a simple helper method:

```go

func (p *PersistenceTx) retrieveDeviceByQuery(query string) (_ *persistence.AuthorizedDevice, ok bool, err error) {
var d persistence.AuthorizedDevice
	err = p.tx.Query(query).WithContext(p.ctx).Scan(&d.Owner, &d.DeviceID)
	if err != nil && errors.Is(err, gocql.ErrNotFound) {
		return nil, false, nil
	}
	if err != nil {
		return nil, false, err
	}
	return &d, true, nil
}

func (p *PersistenceTx) Retrieve(deviceID, owner string) (_ *persistence.AuthorizedDevice, ok bool, err error) {
  ...
  build query
  ...

  return p.retrieveDeviceByQuery(b.String());
}

func (p *PersistenceTx) RetrieveByDevice(deviceID string) (_
  ...
  build query
  ...
  
  return p.retrieveDeviceByQuery(b.String());
}

```

identity-store/persistence/cqldb/persist.go
```diff
@@ -0,0 +1,227 @@
...
+	iter := p.tx.Query(b.String()).WithContext(p.ctx).Iter()
+
+	return &iterator{
+		iter: iter,
+		ctx:  p.ctx,
+	}
+}
+
+// RetrieveAll retrieves all user's authorized devices.
+func (p *PersistenceTx) RetrieveAll() persistence.Iterator {
```
Danielius1922: SC reports that this is never used, can't we add a simple test?

identity-store/persistence/cqldb/persist.go
```diff
@@ -0,0 +1,227 @@
...
+	b.WriteString(deviceIDKey)
+	b.WriteString("=")
+	b.WriteString(deviceID)
+
+	return p.tx.Query(b.String()).WithContext(p.ctx).Exec()
+}
+
+// Close must always be called (use defer immediately after NewTransaction).
+func (p *PersistenceTx) Close() {
+	// do nothing - transaction is not supported
```
Danielius1922: So these are not transactions? Can't that be an issue?

pkg/cqldb/config.go
```diff
@@ -0,0 +1,107 @@
...
+		return fmt.Errorf("tls.%w", err)
+	}
+	if err := c.Keyspace.Validate(); err != nil {
+		return fmt.Errorf("keyspace.%w", err)
+	}
+	if c.Port < 0 || c.Port > 65535 {
+		return fmt.Errorf("port('%v') - invalid port", c.Port)
+	}
+	if c.Port == 0 {
+		c.Port = 9142
```
Danielius1922: put a named constant (`CQLDB_DEFAULT_PORT` maybe?) at the begining of the file

identity-store/persistence/cqldb/store.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func New(ctx context.Context, config *Config, fileWatcher *fsnotify.Watcher, logger log.Logger, tracerProvider trace.TracerProvider) (*Store, error) {
+	certManager, err := client.New(config.Embedded.TLS, fileWatcher, logger)
+	if err != nil {
+		return nil, fmt.Errorf("could not create cert manager: %w", err)
+	}
+	cqldbClient, err := cqldb.New(ctx, config.Embedded, certManager.GetTLSConfig(), logger, tracerProvider)
+	if err != nil {
+		return nil, err
```
Danielius1922, line 48: leak: add `certManager.Close` before returning (same for line 53)

pkg/cqldb/cqldb.go
```diff
@@ -0,0 +1,132 @@
...
+	"strings"
+
+	"github.com/gocql/gocql"
+	"github.com/plgd-dev/hub/v2/pkg/log"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type OnClearFn = func(context.Context) error
+
+// Client implements an Client for MongoDB.
```
Danielius1922: not MongoDB

pkg/cqldb/cqldb_test.go
```diff
@@ -0,0 +1,92 @@
...
+	require.NoError(t, err)
+	fmt.Println(time.Since(now))
+
+	scanner := s.Query("SELECT pk, ck, description FROM example.batches").WithContext(ctx).Iter().Scanner()
+	for scanner.Next() {
+		var pk, ck int32
+		var description string
+		err = scanner.Scan(&pk, &ck, &description)
+		require.NoError(t, err)
+		fmt.Println(pk, ck, description)
```
Danielius1922: maybe at least verify that we get 2 values and check the `description` values (should be `${pk}.${ck}`

resource-aggregate/cqrs/eventstore/cqldb/delete_test.go
```diff
@@ -0,0 +1,159 @@
...
+			},
+			wantErr: false,
+		},
+		{
+			name: "Invalid and valid groupID",
+			args: args{
+				query: []eventstore.DeleteQuery{{
+					GroupID: uuid.Nil.String(),
+				}, {
+					GroupID: getDeviceID(1),
```
Danielius1922, line 103: Is there a simple way to verify that the DB no longer contains any rows with the given `groupID` after this iteration runs?

resource-aggregate/cqrs/eventstore/cqldb/eventstore.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, o := range opts {
+		o.apply(config)
+	}
+	certManager, err := client.New(config.Embedded.TLS, fileWatcher, logger)
+	if err != nil {
+		return nil, fmt.Errorf("could not create cert manager: %w", err)
+	}
+	cqldbClient, err := cqldb.New(ctx, config.Embedded, certManager.GetTLSConfig(), logger, tracerProvider)
+	if err != nil {
+		return nil, err
```
Danielius1922, line 67: leak? `certManager.Close` before line 73 (and 78)

resource-aggregate/cqrs/eventstore/cqldb/delete_test.go
jkralik: There is the code at the end of each test  that gets all events and check if the devices haven't more not exists:

```
			// get all events after deletion
			handler := test.NewMockEventHandler()
			err = store.GetEvents(ctx, []eventstore.GetEventsQuery{{}}, 0, handler)
			require.NoError(t, err)
			// no documents with deleted group id should remain
			for _, q := range tt.args.query {
				require.False(t, handler.ContainsGroupID(q.GroupID))
			}
```

resource-aggregate/cqrs/eventstore/cqldb/eventstore.go
jkralik: Fixed

pkg/cqldb/cqldb_test.go
jkralik: Fixed

pkg/cqldb/cqldb.go
jkralik: Fixed

pkg/cqldb/config.go
jkralik: Fixed

identity-store/persistence/cqldb/persist.go
jkralik: Fixed

identity-store/persistence/cqldb/persist.go
jkralik: Removed

identity-store/persistence/cqldb/persist.go
jkralik: Fixed

certificate-authority/store/cqldb/store.go
jkralik: Fixed

certificate-authority/store/cqldb/store.go
jkralik: Fixed

certificate-authority/store/cqldb/store.go
jkralik: Fixed

certificate-authority/store/cqldb/signingRecords_test.go
jkralik: Removed

certificate-authority/store/cqldb/signingRecords.go
jkralik: Fixed

resource-aggregate/cqrs/eventstore/cqldb/load.go
```diff
@@ -0,0 +1,228 @@
...
+		return nil, false
+	}
+	version, ok := ev[versionKey].(int64)
+	if !ok {
+		i.err = fmt.Errorf("invalid data, '%v' of event %v is not an int64 but %T", versionKey, eventTypeKey, ev[versionKey])
+		return nil, false
+	}
+	aggregateID, ok := ev[idKey].(gocql.UUID)
+	if !ok {
+		i.err = fmt.Errorf("invalid data, '%v' of event %v is not a string but %T", idKey, eventTypeKey, ev[idKey])
```
Danielius1922: `is not a string` -> `is not a gocql.UUID`?

resource-aggregate/cqrs/eventstore/cqldb/load.go
```diff
@@ -0,0 +1,228 @@
...
+		i.err = fmt.Errorf("invalid data, '%v' of event %v is not a string but %T", deviceIDKey, eventTypeKey, ev[deviceIDKey])
+		return nil, false
+	}
+	snapshot, ok := ev[snapshotKey].([]byte)
+	if !ok {
+		i.err = fmt.Errorf("invalid data, '%v' of event %v is not a []byte but %T", snapshotKey, eventTypeKey, ev[snapshotKey])
+		return nil, false
+	}
+	return eventstore.NewLoadedEvent(
+		uint64(version),
```
Danielius1922, line 85: why the first conversion to `int64` on line 61 and then this second cast? can't it be `uint64` from the start?

resource-aggregate/cqrs/eventstore/cqldb/load.go
```diff
@@ -0,0 +1,228 @@
...
+
+	filter := snapshotQueriesToFilter(groupID, queries, timestamp)
+	return s.loadEventsQuery(ctx, eventHandler, filter)
+}
+
+func normalizeQueries(queries []eventstore.SnapshotQuery) map[string][]eventstore.SnapshotQuery {
+	normalizeQuery := make(map[string][]eventstore.SnapshotQuery, len(queries))
+	for _, query := range queries {
+		if query.AggregateID == "" {
+			normalizeQuery[query.GroupID] = make([]eventstore.SnapshotQuery, 0, 1)
```
Danielius1922: make([]eventstore.SnapshotQuery) doesn't seem like the 1 capacity is needed

resource-aggregate/cqrs/eventstore/cqldb/loadDeviceMetadataByServiceIDs.go
```diff
@@ -0,0 +1,55 @@
...
+	if iter == nil {
+		return nil, errors.New("cannot create iterator")
+	}
+	ret := make([]eventstore.DeviceDocumentMetadata, 0, iter.NumRows())
+	values := make(map[string]interface{}, 2)
+	for iter.MapScan(values) {
+		deviceID, ok := values[deviceIDKey].(gocql.UUID)
+		if !ok {
+			_ = iter.Close()
+			return nil, fmt.Errorf(errFmtDataIsNotStringType, deviceIDKey, values[deviceIDKey])
```
Danielius1922: errFmtDataIsNotStringType -> is not a `gocql.UUID`

resource-aggregate/cqrs/eventstore/cqldb/maintenance.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Query retrieves the latest snapshot version per aggregate for thw number of aggregates specified by 'limit'
+func (s *EventStore) Query(_ context.Context, _ int, _ maintenance.TaskHandler) error {
+	return fmt.Errorf("not implemented")
+}
+
+// Remove deletes (the latest snapshot version) database record for a given aggregate ID
+func (s *EventStore) Remove(_ context.Context, _ maintenance.Task) error {
+	// not supported
```
Danielius1922, line 17: `return eventstore.ErrNotSupported`?

certificate-authority/store/cqldb/signingRecords.go
```diff
@@ -32,16 +32,14 @@ func setTTLbyDeviceID(deviceID string, validUntil int64) string {
...
 	}
-
 	data, err := utils.Marshal(signingRecord)
 	if err != nil {
-		return err
+		return "", err
 	}
-
 	var b strings.Builder
 	b.WriteString("INSERT INTO ")
```
Danielius1922, line 44: must be `UPDATE ${table} SET` instead of `INSERT INTO ${table}`

identity-store/persistence/cqldb/store.go
```diff
@@ -84,6 +87,14 @@ func newEventStoreWithClient(ctx context.Context, client *cqldb.Client, config *
 		return nil, err
 	}
 
+	// TODO remove me
+	time.Sleep(time.Millisecond * 100)
```
Danielius1922: rm?

resource-aggregate/cqrs/eventstore/cqldb/load.go
jkralik: No because, cqldb doesn't support `uint64`

certificate-authority/store/cqldb/signingRecords.go
jkralik: Why ? `INSERT INTO ` in cql works as upsert in some cases and we use it to set TTL for all colums

identity-store/persistence/cqldb/store.go
jkralik: removed

resource-aggregate/cqrs/eventstore/cqldb/loadDeviceMetadataByServiceIDs.go
jkralik: fixed

resource-aggregate/cqrs/eventstore/cqldb/maintenance.go
jkralik: fixed

resource-aggregate/cqrs/eventstore/cqldb/load.go
jkralik: fixed

resource-aggregate/cqrs/eventstore/cqldb/load.go
jkralik: fixed

resource-aggregate/cqrs/eventstore/cqldb/save.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return b.String()
+}
+
+func eventToKeyValues(event eventstore.Event, insert bool, data []byte) keyValues {
+	kvs := make([]keyValue, 0, 16)
+	if insert {
+		kvs = append(kvs, keyValue{Key: deviceIDKey, Val: event.GroupID()})
+		kvs = append(kvs, keyValue{Key: idKey, Val: event.AggregateID()})
+		kvs = append(kvs, keyValue{Key: eventTypeKey, Val: "'" + event.EventType() + "'"})
```
Danielius1922, line 55: Is there a reason why `'` is used for eventType but not for deviceID and id?

pkg/cqldb/hostDialer.go
```diff
@@ -0,0 +1,100 @@
+package cqldb
```
Danielius1922, line 1: This file has 0% code coverage, can we add some tests?

resource-aggregate/cqrs/eventstore/cqldb/save.go
jkralik: in cql you need to set text values with `'`. deviceIDKey and idKey they are cql UUID types so and the end it is converted to 16bytes.

pkg/cqldb/hostDialer.go
jkralik: Fixed

certificate-authority/config.yaml
```diff
@@ -68,6 +69,28 @@ clients:
         timeout: 1m0s
         throttleTime: 500ms
         documentLimit: 1000
+    cqlDB:
```
ondrejtomcik, line 72: configuration of the database can differ even though their query language is the same. Replication factor, numConnections, and other properties might or might not exist or use different names. Therefore I would propose to use directly scyllaDB as a key.

certificate-authority/store/cqldb/config.go
```diff
@@ -0,0 +1,15 @@
+package cqldb
```
ondrejtomcik, line 1: also here + folder name. This might really differ per db implementing CQL

pkg/config/database/config.go
```diff
@@ -0,0 +1,52 @@
...
+	MongoDB DBUse = "mongoDB"
+	CqlDB   DBUse = "cqlDB"
+)
+
+type DBConfig interface {
+	Validate() error
+}
+
+type Config[MongoConfig DBConfig, CQLDBConfig DBConfig] struct {
+	Use     DBUse       `yaml:"use" json:"use"`
```
ondrejtomcik, line 25: Do we need this struct at all? Why not to skip it completely, and work only with - if url is set on mongo db, if not, if url is set on scyllaDB, and decide based on this. What do you think?

pkg/cqldb/config.go
```diff
@@ -0,0 +1,107 @@
...
+		return fmt.Errorf("constant.%w", err)
+	}
+	return nil
+}
+
+type Config struct {
+	Hosts                 []string                 `yaml:"hosts" json:"hosts"`
+	Port                  int                      `yaml:"port" json:"port"`
+	NumConns              int                      `yaml:"numConnections" json:"numConnections"`
+	ConnectTimeout        time.Duration            `yaml:"connectTimeout" json:"connectTimeout"`
```
ondrejtomcik, line 77: Shouldn't this be `ConnectionTimeout`? Why don't we have this prop also for mongodb client?

ondrejtomcik, line 77: Isn't this the same as MaxConIdleTime for mongodb?

pkg/cqldb/config.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := c.Constant.Validate(); err != nil {
+		return fmt.Errorf("constant.%w", err)
+	}
+	return nil
+}
+
+type Config struct {
+	Hosts                 []string                 `yaml:"hosts" json:"hosts"`
+	Port                  int                      `yaml:"port" json:"port"`
+	NumConns              int                      `yaml:"numConnections" json:"numConnections"`
```
ondrejtomcik, line 76: Isn't this equivalent of maxPoolSize we have in mongodb?

pkg/cqldb/config.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (c *ReconnectionPolicyConfig) Validate() error {
+	if err := c.Constant.Validate(); err != nil {
+		return fmt.Errorf("constant.%w", err)
+	}
+	return nil
+}
+
+type Config struct {
+	Hosts                 []string                 `yaml:"hosts" json:"hosts"`
```
ondrejtomcik, line 74: is host a scylladb driver terminology? Asking because in mongodb the terminologi is uri

certificate-authority/config.yaml
jkralik: The part `cqldb.Keyspace.Replication` is generic so it will forward all values to cqlDB. All other configs are same for both databases scylladb/apache cassandra.

certificate-authority/store/cqldb/config.go
jkralik: This is a client that you can use with CQL databases.

pkg/config/database/config.go
jkralik: Because both part could be configured and with this option you can simple change DB. Yes it is more helpful struct for future.

pkg/cqldb/config.go
jkralik: The parameter said when the connect cannot be establish to DB. When client is not able connect to DB then the service need to throw error and ends.

jkralik: No. This is about how long it wil try to create connection. If the connectTimeout is reached then it will report a error.

pkg/cqldb/config.go
jkralik: Yes and no. NumConns really creates number of the connection - but the names of properties follows cql client configuration.

pkg/cqldb/config.go
jkralik: Yes. It is. It contains just hosts without port and schemes.

pkg/cqldb/config.go
ondrejtomcik: But why don't we have this prop also for mongodb client?

ondrejtomcik: Is this terminology from cql client?

pkg/config/database/config.go
ondrejtomcik: I don't see a use case to configure both databases. There is no real world scenario I can come up with. You?

jkralik: You need to choose which you want to use it. It is much better when you dont have have property with two meanings (configured = host )like URL/HOST. 
Eg migration - if we support it in future then you need to have congifured both values + direction.

ondrejtomcik (approved):

PR #1122, opened 2023-11-06
Fix tests for new x.org.iotivity.version property in the platform resource

Danielius1922:


test/test.go
```diff
@@ -142,6 +142,8 @@ var (
 
 	TestDevsimResources        []schema.ResourceLink
 	TestDevsimBackendResources []schema.ResourceLink
+
+	testIovityLiteVersion map[string]uint32
```
jkralik: sync.Map ? I think it could be better for future

jkralik (approved):

PR #1137, opened 2023-11-15
Helm: Customizable UI Theme for Customers

jkralik:
This pull request introduces a significant enhancement to our web application by enabling customers to personalize their user interface (UI) theme. Users can now tailor the visual appearance of the application according to their preferences, fostering a more personalized and engaging user experience.


charts/plgd-hub/templates/_helpers.tpl
```diff
@@ -162,9 +162,9 @@ If release name contains chart name it will be used as a full name.
     {{- $raClientTls := .Values.coapgateway.clients.resourceAggregate.grpc.tls.certFile }}
     {{- $rdClientTls := .Values.coapgateway.clients.resourceDirectory.grpc.tls.certFile }}
     {{- if and $natsTls $authClientTls $raClientTls $rdClientTls }}
-    {{- printf "false" }}
+    {{- printf "" -}}
```
Danielius1922, line 165: why the "-" at the end "-}}", all other lines do not have it, does it mean something or it is optional?

EDIT: ok, chatGPT says that it means that no trailing whitespace will be printed when the last "-" is used

Danielius1922 (approved):

PR #1225, opened 2024-02-02
snapshot-service: Introduce plgd/snapshot-service

jkralik:
The PLGD Snapshot Service is designed for managing devices snapshots. It provides functionality for creating, retrieving,  deleting snapshots, applying snapshots, managing snapshot associations, and obtaining snapshot statuses.


snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+}
+
+message Configuration {
+  message Resource {
+    oneof filter {
+      string href  = 1;
+      ResourceTypes resource_types = 2; 
+    }
+    resourceaggregate.pb.Content content = 3;
+    string resource_interface = 6; // optional update interface
```
ondrejtomcik: @jkralik I would skip support for resource interface for now. We can add it later. I assume nobody is using this feature, so in case we decide in future to deprecate it completely - plgd device, we don't have to introduce breaking change here.

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+    int64 time_to_live = 7; // optional update command time to live
+  }
+  Id id = 1;
```
ondrejtomcik: @jkralik It might be misleading, as we usually use UUIDs for IDs in plgd hub. I would propose to not encapsulate version into the ID. What about having id and version directly in configuration struct?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+  string name = 2;
+  repeated Resource resources = 3;
+}
+
+message CreateConfigurationReponse { Id id = 1; }
```
ondrejtomcik: @jkralik In our API calls, we tend to return created/updated content. Why not here?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+message UpdateConfigurationReponse  {
+  Id id = 1; // for new version 
+} 
+message GetConfigurationsRequest { repeated IdFilter id_filter = 1; }
+message DeleteConfigurationsRequest { repeated IdFilter id_filter = 1; }
+message DeleteConfigurationsResponse { int64 count = 1; }
```
ondrejtomcik, line 84: @jkralik it's better to return `repeated string configuration_ids = 1;` in order to inform user what was not deleted and re-render UI without a need to do a GET.

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+
+message AppliedConfiguration {
+  message Resource {
+    resourceaggregate.pb.ResourceId resource_id = 1;
+    uint32 configuration_resources_idx = 2; // index of resource in configuration resources array. For resource types it could be mutliple resources.
+    enum Status {
```
ondrejtomcik, line 90: @jkralik what is `WAITING_FOR_RESOURCE` state describing? It would be good to have in system compatible states, mainly if they represent contextually same behaviour. Why not to reuse pending commands states?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+message CreateConfigurationReponse { Id id = 1; }
+message UpdateConfigurationReponse  {
+  Id id = 1; // for new version 
+} 
+message GetConfigurationsRequest { repeated IdFilter id_filter = 1; }
+message DeleteConfigurationsRequest { repeated IdFilter id_filter = 1; }
+message DeleteConfigurationsResponse { int64 count = 1; }
+
+message AppliedConfiguration {
+  message Resource {
```
ondrejtomcik, line 87: @jkralik wouldn't it be better idea to reuse [pending commands](https://github.com/plgd-dev/hub/blob/main/grpc-gateway/pb/getPendingCommands.proto#L29C9-L29C51) for this?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+    };
+    Status status = 3;
+    int64 timestamp_start = 4; // when the rule association was applied
+    int64 valid_until = 5; // how long the command is valid
+    resourceaggregate.pb.ResourceUpdated resource_updated = 6;
+  }
+  string id = 1;
+  string device_id = 2;
+  Id configuration_id = 3;
+  oneof executed_by {
```
ondrejtomcik, line 106: @jkralik what if the same configuration id+version is executed by multiple conditions and also on demand? Will this result in multiple AppliedConfigurations?

What if the trigger has the same rule? -> How to evaluate if update was executed already (not saying we need to do so)?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+    bool on_demaned = 4;
+    Id condition_id = 5;
+  }
+  repeated Resource resources = 6;
+}
+
+message InvokeConfigurationRequest {
+  string id = 1;
+  repeated string device_id_filter = 2; // at least one must be set
```
ondrejtomcik: @jkralik Will invocation result in one `AppliedConfiguration` with resources from e.g. 3 devices? Or is the granularity of the `AppliedConfiguration` per Device?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+  Id configuration_id = 3;
+  oneof executed_by {
+    bool on_demaned = 4;
+    Id condition_id = 5;
+  }
+  repeated Resource resources = 6;
+}
+
+message InvokeConfigurationRequest {
+  string id = 1;
```
ondrejtomcik: @jkralik why do you need identifier for invocation? Do you plan to persist them?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,265 @@
...
+  repeated string device_id_filter = 2; // at least one must be set
+  bool force = 3;
+  bool keep_updating_on_failure = 4;
```
ondrejtomcik: @jkralik could you describe scenarios for both, force and keep_updating please?

snapshot-service/pb/service.proto
jkralik: ok - removed

snapshot-service/pb/service.proto
jkralik: MongoDB doesn't provide such functionality -> the only way is that you do delete them one by one ...

snapshot-service/pb/service.proto
jkralik: When resource still not exists in the hub, it is wait for him - it is expected that the resource will be created in short time.

snapshot-service/pb/service.proto
jkralik: It is expected that the applied configuration for `device_id, id` only once.

snapshot-service/pb/service.proto
jkralik: - force: re-apply configuration to device
- keep_updating_on_failure: when some update resource fails the updates will continue (maybe this flag could be moved to Resource Config so for each update resource you can said if you want to continue)

snapshot-service/pb/service.proto
jkralik: Will invoke set configuration with id to 3 devices.

snapshot-service/pb/service.proto
jkralik: No - it is different. Status contains other enums.

snapshot-service/pb/service.proto
jkralik: updated

snapshot-service/pb/service.proto
jkralik: updated

snapshot-service/pb/service.proto
jkralik: ID says which configuration you wan to apply to the devices in device_id_filter.

snapshot-service/pb/service.proto
ondrejtomcik: @jkralik in case of devices, I know there our APIs are involved, it's possible. Can we somehow support it as well, by doing a db.read after delete and then returning? UI will anyway need to call GET if you don't provide it in the response.

snapshot-service/pb/service.proto
ondrejtomcik: Should it be different? Aren't we just creating pending commands?

snapshot-service/pb/service.proto
ondrejtomcik: That's equal to Pending Command. Isn't it? And timeout is expired for example. Aren't we dealing only with pending commands and their result state?

snapshot-service/pb/service.proto
jkralik: The only possible way is to delete one by one. other ways are inefficient.

snapshot-service/pb/service.proto
ondrejtomcik: Do you mean `device_id` + `cfg id` + `cfg version`?

But how do you want to prevent that Configuration of the same id and version is not applied twice on the same device, if there are two rules referencing it, or it's invoked manually?

snapshot-service/pb/service.proto
```diff
@@ -0,0 +1,257 @@
...
+    enum Status {
+      QUEUED = 0;
+      INPROGRESS = 1;
+      WAITING_FOR_REOURCE = 2;
+      DONE = 3;
+      TIMEOUT = 4;
+      FAIL = 5;
+    };
+    Status status = 3;
+    int64 timestamp_start = 4; // when the rule association was applied
```
ondrejtomcik: @jkralik how this differs from the applied? If the device connects, that has serial number 123, rule is triggered and immediately the pending command is created. What this timestamp start really represents? Connected device? Resource on which rule depends published and retrieved? Command queued? Command done/fail?

snapshot-service/pb/service.proto
ondrejtomcik: Then use `configuration_id` instead please.

snapshot-service/pb/service.proto
ondrejtomcik: That's not what I asked. This requests accepts filter, I can specify 3 devices. Applied configuration, which represents what happened, has a device id as a single string. That means, if 3 devices are found from the filter specified in this request, it will result in 3 `AppliedConfiguration`, each with a device id matching the filter device ids. Right?

snapshot-service/pb/service.proto
jkralik: No it's different. Currently you cannot create update resource when the resource changed of the resource is not stored  or it is not published. We can introduce `bool flag` which allows to create update pending request even the resource not exist. If update resource command to aggregate fails we need to somehow report the error.

snapshot-service/pb/service.proto
jkralik: No - just `device_id` + `cfg id`. This means that the configuration has been applied. In this case version is used only to find the configuration which has been applied.

snapshot-service/pb/service.proto
jkralik: `InvokeConfigurationRequest.configuration_id` it seems to me that i repeat to often the word `configuration`.

snapshot-service/pb/service.proto
jkralik: The configuration is applied one by one - Not parallel (maybe in future we can add such option).
`timestamp_start` it is a time when the resource update command to resource aggregate will be send.

snapshot-service/pb/service.proto
jkralik: Right

PR #1229, opened 2024-02-29
coap-gateway/resource-aggregate: add device local endpoints support

jkralik:
Obtain the local endpoints of a device within a Local Area Network (LAN) that is situated behind a Network Address Translation (NAT) firewall from the hub.


resource-aggregate/events/deviceMetadataUpdated.go
```diff
@@ -69,11 +69,24 @@ func (d *DeviceMetadataUpdated) CheckInitialized() bool {
 		d.GetEventMetadata() != nil
 }
 
+func equal[T comparable](a, b []T) bool {
```
Danielius1922: slices.Equal (https://pkg.go.dev/slices#Equal)?

Danielius1922 (approved):

resource-aggregate/events/deviceMetadataUpdated.go
jkralik: for now ext package -> in 1.22 it is supported

PR #1348, opened 2024-07-12
snippet-service: rename resourceUpdater config to resourceAggregate

jkralik:


charts/plgd-hub/values.yaml
```diff
@@ -2527,7 +2527,7 @@ snippetservice:
           keyFile:
           certFile:
           useSystemCAPool: false
-    resourceUpdater:
+    resourceAggregate:
       cleanUpExpiredUpdates: "0 * * * *"
```
Danielius1922: This is a resourceUpdater setting though. I guess we should have both? ResourceAggregate and ResourceUpdater configs?

jkralik: this is a client to `resourceAggregate` -> what you want to move to `ResourceUpdater` ? just `cleanUpExpiredUpdates: "0 * * * *" ?`

Danielius1922: `CleanUpExpiredUpdates` and `ExtendCronParserBySeconds` should be outside of `resourceAggregate`. I guess we can move them to the root level if we want to avoid making a new resourceUpdaterConfig struct?

jkralik: We have something similar here:
image

https://github.com/plgd-dev/hub/blob/5f8588a972a8b95709f8af8266e4095b6fc82db7/charts/plgd-hub/values.yaml#L2247

Maybe we can leave it as it is. What do you thing?

jkralik: moved to storage config

Danielius1922 (approved):

PR #1352, opened 2024-07-23
Check whether token is trusted after validation

Danielius1922:


pkg/security/jwt/cache.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func NewTokenCache() *TokenCache {
+	return &TokenCache{
+		cache: cache.NewCache[string, TokenRecord](),
+	}
+}
+
+func getCacheKey(tokenID, issuer string) string {
+	return tokenID + "." + issuer
```
jkralik: this could be very big keys ~ 69bytes 37(uuid) + ~32 (https://{host}/m2m-oauth-server)

pkg/security/jwt/validator.go
```diff
@@ -34,6 +90,82 @@ func errParseTokenInvalidClaimsType(t *jwt.Token) error {
...
+
+	tr, ok := v.tokenCache.Load(tokenID, issuer)
+	if ok {
+		if tr.Blacklisted {
+			return ErrBlackListedToken
+		}
+		return nil
+	}
+
+	uri, err := url.Parse(client.tokenEndpoint)
```
jkralik: until the token is not in cache you will make each request even the token is same. Make only one request for one token in same time.

pkg/security/jwt/tokenCache.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+type tokenIssuerCache struct {
+	client        *http.Client
+	tokenEndpoint string
+	tokens        map[uuid.UUID]tokenOrFuture
+	mutex         sync.Mutex
+}
+
+func newTokenIssuerCache(client *Client) *tokenIssuerCache {
```
jkralik, line 84: I dont see the code for cleanup records from the cache

Danielius1922: wait 2 mins

jkralik (approved):

